PR #346, opened 2014-08-02
JDK 8u20 for Yosemite

dsbaars:
To use eclipse in Mac OS X 10.10 you need this early access release, as pointed out by an Oracle employee at https://community.oracle.com/thread/3568196


radeksimko:
It won't be as easy as I thought... closing this in favour of #831

PR #640, opened 2015-01-01
Add World of Tanks Latest

shpoont:


vitorgalvao:
The `-na` version should be removed, as it’d be a duplicate of the one to be used as default.

PR #642, opened 2015-01-02
Add PhpStorm EAP v139.873 Cask

zchee:


vitorgalvao:
I was under the impression this [wasn’t working anymore](https://github.com/caskroom/homebrew-versions/pull/621). Have they reversed the position?

Pining @vinkla.

vinkla:
Yes, [the EAP links](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Early+Access+Program) are now back. 

There were a gap between december 11 and 24 when no downloadable links were to be found. The old links were broken and unusable. It seems that they remove the EAP version when it goes stable and only creates a new EAP when its ready for the public. This may occur again in the future. 

Should we maybe keep them out since we can't predict this? 

[Read more information on their blog.](http://blog.jetbrains.com/phpstorm/)

vitorgalvao:
> Should we maybe keep PhpStorm and WebStorm out since we can't predict this?

Possibly, yes, but as I don’t use these I have no idea what would be preferable for its users. Merging this one since it’s functional again, then.

PR #654, opened 2015-01-08
Add FoldingText dev version

christopherdwhite:
Added the development version of FoldingText to as a new cask. FoldingText dev versions use a single download link so this cask will always be up to date unless the download URL is changed.

### DIFF
--- a/Casks/foldingtext-dev.rb
+++ b/Casks/foldingtext-dev.rb
@@ -1,0 +1,12 @@
+cask :v1 => 'foldingtext-dev' do
+  version :latest
+  sha256 :no_check
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
You still need to point me to the download page, so I can verify the link.

christopherdwhite:
Sorry, it's been a long day.

Here's the download page for the current dev version:

http://support.foldingtext.com/t/foldingtext-2-1-dev-741/382

It looks like Jesse creates new threads for new versions but always uses the same download link.

vitorgalvao:
Thank you.

PR #673, opened 2015-01-15
Add io.js 1.0.1

ingorichter:
This will replace node.js with a better node.js aka iojs

### DIFF
--- a/Casks/iojs.rb
+++ b/Casks/iojs.rb
@@ -1,0 +1,16 @@
+cask :v1 => 'iojs' do
+  version 'v1.0.1'
+  sha256 'af3fb3dfbd7e21db007d29610d1337b5cb6ec55fcb7e8d3240471059dd123f27'
+
+  url "https://iojs.org/dist/#{version}/iojs-#{version}.pkg"
+  name 'io.js'
+  homepage 'https://iojs.org'
+  license :oss
+
+  pkg "iojs-#{version}.pkg"
+
+  uninstall :pkgutil => [
+                         'com.nodesource.iojs.npm.pkg',
+                         'com.nodesource.iojs.pkg'
+                        ]
+end


MoOx:
Be careful, another PR of iojs has been submitted directly to homebrew https://github.com/Homebrew/homebrew/pull/35853

vitorgalvao:
@MoOx Different projects. This is still perfectly valid, here.

Our policies are different, though. That version of `iojs` is the best one available, so [independently of what upstream is calling it](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#but-there-is-no-stable-version), it still belongs in the [main repo](https://github.com/caskroom/homebrew-cask). @ingorichter I as that you submit this there.

ingorichter:
I'm not quite sure about submitting to the main repo. Since there is already https://github.com/Homebrew/homebrew/pull/35853 waiting to be merged. I though, that this solution is different in the way, that it uses the installer instead of building everything from source. I don't know how this is usually handled, if there are multiple ways to install an app.

vitorgalvao:
@ingorichter Like I said, this is a different project; it isn’t unusual for us to share software with Homebrew (see [jenkins](https://github.com/caskroom/homebrew-cask/blob/202845ca8a7f58231e4240a96e61e04fc943bdfe/Casks/jenkins.rb) or [macvim](https://github.com/caskroom/homebrew-cask/blob/eb4f8e1882ef99273a197fb0f30797ce98e76288/Casks/macvim.rb)).

> I don't know how this is usually handled

Like I told you, by submitting to the main repo. I’ll reiterate we’re different projects: we don’t tend to include or exclude things based of each other.

Please submit the cask to the main repo. If you don’t want to, please let me know and someone else will do it.

ingorichter:
Okay, thanks for clarifying. I'll submit it in main repo.

vitorgalvao:
Thank you. You can add a `conflicts_with :formula => 'iojs'` line near the end, if you want to. The option isn’t yet functional, but it can be included without repercussions.

PR #676, opened 2015-01-15
Added Raindrop.io beta 3.3.4

lucasmotta:


vitorgalvao:
Thank you, but [this belongs in the main repo](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#but-there-is-no-stable-version), and has already been [been submitted there](https://github.com/caskroom/homebrew-cask/pull/8935/).

Some corrections, for future submissions, though. Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/master/Casks/acorn.rb). `homepage` should have a trailing `/` in this case. If license is left as `:unknown`, it [requires a comment in front](https://github.com/caskroom/homebrew-cask/blob/4ef9d45417eefaa6510294a999dff34c7f1d7b01/Casks/miro.rb#L8).

PR #681, opened 2015-01-16
Added Earlybird 37.0a2 aka Thunderbird

zerrot:
A better cask token is probably thunderbird-nightly or thunderbird-earlybird, to show the relationship to Thunderbird, but to keep with the contribution guidelines I created the token based on the .app name.
I included Thunderbird Nightly as an alternative name so it is clear what the app relationship is when looking at the info.


vitorgalvao:
Nicely done. You had to deduce some things, but made the correct decisions at every step. Thank you.

zerrot:
Thanks!

PR #683, opened 2015-01-16
XMLMind Editor 3.5.2

zzamboni:
Cask for XMLMind Editor 3.5.2, which is an old version but the last one that allows commercial use without purchasing a license.

### DIFF
--- a/Casks/xmlmind-editor352.rb
+++ b/Casks/xmlmind-editor352.rb
@@ -7,7 +7,13 @@ cask :v1 => 'xmlmind-editor352' do
   homepage 'http://www.xmlmind.com/xmleditor/'
   license :commercial
 
-  depends_on :cask => 'java6'
-
   app 'xxe-std-3_5_2/XMLEditor.app'
+
+  caveats <<-EOS.undent
+    #{token} may require Java 6 (an older version), available from the
+    caskroom-versions repository via
+
+      brew cask install caskroom/versions/java6
+
+  EOS
 end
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Are you sure this only works with `java6` and not others? Either way, that’s not the correct way to do it (java is a complex issue), and we have a [`depends_on :java`](https://github.com/caskroom/homebrew-cask/issues/7993) in the works (not yet functional).

A `caveat` is currently the correct solution for this. Which one depends on if this only works with that version of java or not.

Also pinging @radeksimko on this, as I do with many java-related issues.

zzamboni:
Thanks for the feedback. The app does not work with the system Java in Yosemite (1.8.0_25), produces some class conversion errors. With java6 it works fine. I have not tested with java7.

Can you suggest what would be the appropriate caveats stanza in this case? Something like this maybe?

    caveats “If XMLEditor does not open files correctly, try installing the java6 cask”

zzamboni:
I have replaced the java6 dependency with a caveats stanza. In my testing this application does not work with Java 7 or 8, it needs Java 6.

vitorgalvao:
Thank you for the contribution. It will be merged in [a different pull request](https://github.com/caskroom/homebrew-cask/pull/696) due to some simple recommended alterations. Your contribution is still included (and still credited to you), with the appropriate modifications, when applicable.

Changes:
+ [Facilitated future updates](https://github.com/caskroom/homebrew-cask/pull/4910).
+ [Squashed commits](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#squashing).
+ Removed extraneous line in caveat.

zzamboni:
Thanks!

PR #688, opened 2015-01-20
rstudio-preview: updates to support nightly

aguynamedryan:
RStudio Preview is updated several times a week.

Instead of submitting an updated formula for every update, I've
modified the formula to mimic the iterm-nightly formula.

Unfortunately, RStudio doesn't (yet) provide a URL that points
to the latest RStudio Preview download, so I've created a URL
through tiny.cc that I will update when new versions of RStudio
Preview are released.


vitorgalvao:
If you’re hosting this yourself, we can’t accept it, as we only take official urls. There are [a few reasons](https://github.com/caskroom/homebrew-cask/pull/615#issuecomment-19972408), the most important being that users should be able to trust the source of the download.

aguynamedryan:
Understood.  I'll go badger the RStudio guys.  Thanks.

PR #689, opened 2015-01-21
Adding Chirp Daily Builds

mannkind:
In case anyone wants the daily build of chirp. Sadly the filename is never the same so I had to find it within the latest SHA1SUM file.

### DIFF
--- a/Casks/chirp-daily-build.rb
+++ b/Casks/chirp-daily-build.rb
@@ -1,0 +1,27 @@
+cask :v1 => 'chirp-daily-build' do
+  module Utils
+    def self.latest_chirp_version
+      # Let's get some info from the SHA1SUM file
+      require 'net/http'
+      html = Net::HTTP.get 'trac.chirp.danplanet.com', '/chirp_daily/LATEST/SHA1SUM'
+      matches = /([a-z0-9]{40})\s+chirp-daily-([0-9]{8})\.app\.zip/.match html
+
+      matches[2]
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Automated solutions like this [have been tried](https://github.com/caskroom/homebrew-versions/pull/293#issuecomment-46122949). However, they were removed due to [technical reasons](https://github.com/caskroom/homebrew-versions/pull/293#issuecomment-48755213). Not sure how your solution would be or not affected by that, so I’ll leave this one to @rolandwalker.

mannkind:
No problem. Chirp (esp. the latest build version) is probably not used by too many people anyway; I just found it useful and thought I might share.

vitorgalvao:
I’d still like to keep it open until @rolandwalker takes a loo. It’s a good contribution, and since I’m not sure this would suffer from the same problem as the other ones and he likely does, if it works it’d be something to consider for other casks.

rolandwalker:
Hi!

Yes this is a problem, arguably because of an architectural failing on our part.  Many operations cause all Casks to be loaded, *ie* fully evaluated.  Thus the HTTP connection here is made when operating on a different Cask.  This will slow things down and sometimes cause wedges.

We are moving towards fixing the architecture, at least for some cases.

I dearly love a hack, and thus want to include this gem, but sadly we cannot handle it.

vitorgalvao:
Closing, then. I’m guessing we won’t forget about adding these when time comes.

PR #703, opened 2015-01-24
Add Scansnap Manager v3.2L31

mbrgm:
This is for Fujitsu ScanSnap S1300.

### DIFF
--- a/Casks/scansnap-manager-sv600.rb
+++ b/Casks/scansnap-manager-sv600.rb
@@ -1,0 +1,15 @@
+cask :v1 => 'scansnap-manager-sv600' do
+  version '6.2L10'
+  sha256 '7273034398e9a57eb0fa89167c9e801ad2bf9fe56b52b3d9591628e978168afb'
+
+  url "https://www.fujitsu.com/downloads/IMAGE/driver/ss/mgr/m-sv600/MacScanSnapV#{version.gsub('.','')}WW.dmg"
+  name "ScanSnap Manager for Fujitsu ScanSnap SV600"
+  homepage 'http://www.fujitsu.com/global/support/computing/peripheral/scanners/software/'
+  license :commercial
+
+  pkg 'ScanSnap Manager.pkg'
+
+  uninstall :pkgutil => 'jp.co.pfu.ScanSnap.*'
+
+  depends_on :macos => '>= :lion'
+end
--- a/Casks/scansnap-manager-sv600.rb
+++ b/Casks/scansnap-manager-sv600.rb
@@ -5,7 +5,7 @@ cask :v1 => 'scansnap-manager-sv600' do
   url "https://www.fujitsu.com/downloads/IMAGE/driver/ss/mgr/m-sv600/MacScanSnapV#{version.gsub('.','')}WW.dmg"
   name "ScanSnap Manager for Fujitsu ScanSnap SV600"
   homepage 'http://www.fujitsu.com/global/support/computing/peripheral/scanners/software/'
-  license :commercial
+  license :gratis
 
   pkg 'ScanSnap Manager.pkg'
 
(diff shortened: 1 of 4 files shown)


Casks/scansnap-manager-v32l31.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  url 'http://www.fujitsu.com/downloads/IMAGE/driver/ss/mgr/m-s1300/ScanSnap.dmg'
+  name 'ScanSnap Manager for Mac (V3.2L31)'
+  homepage 'http://www.fujitsu.com/global/support/computing/peripheral/scanners/software/s1300m-setup.html'
+  license :commercial
+
+  pkg 'Scansnap Manager.pkg'
+  
+  uninstall :pkgutil => 'jp.co.pfu.ScanSnap.V10L10'
+
```
vitorgalvao: Please remove this empty line.

vitorgalvao:
We don’t usually add a `-v…` to versions in this cask, and just right it (so instead of `scansnap-manager-v32l31`, it would be `scansnap-manager32l31`.

In addition, this ends up being somewhat related to https://github.com/caskroom/homebrew-cask/issues/8938. Would you and @bronson perhaps be interested in tackling the rest of them?

mbrgm:
Ok. I changed the cask names. For the slightly older models, this is version/naming hell. I'll remove the unversionend `scansnap-manager` from the regular casks. @bronson mentioned he would be willing to add S1300i as well.

Casks/scansnap-manager-sv600.rb
```diff
@@ -0,0 +1,13 @@
+cask :v1 => 'scansnap-manager-sv600' do
+  version '6.2L10'
+  sha256 '7273034398e9a57eb0fa89167c9e801ad2bf9fe56b52b3d9591628e978168afb'
+
+  url "https://www.fujitsu.com/downloads/IMAGE/driver/ss/mgr/m-sv600/MacScanSnapV#{version.gsub('.','')}WW.dmg"
+  name "ScanSnap Manager for Fujitsu ScanSnap SV600 (OS X 10.7+)"
+  homepage 'http://www.fujitsu.com/global/support/computing/peripheral/scanners/software/'
+  license :proprietary
```
vitorgalvao: Not a [valid license](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#valid-licenses).

mbrgm:
Lol, I always get confused the names for proprietary licenses. In npm I use proprietary, here it is commercial. Confusing :laughing:

vitorgalvao:
I think all the `license` stanzas may be wrong. These are proprietary, yet free to get and use, correct? If that is the case, they should be `:gratis`, not `:commercial`.

The ` (OS X 10.7+)` (and equivalent) should be removed from the `name`s. We have a way better way to do it, [`depends_on :macos`](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#depends_on-macos). We prefer release names, so in the case of `10.7+` it’d be `depends_on :macos => '>= :lion'`. This should be the last line (before the `end`).

vitorgalvao:
Everything is starting to come together nicely; thank you for all your work.

`scansnap-manager-s300m-s510m.rb` is still missing the `depends_on :macos` stanza (and the removal of the version from `name`). Also per my previous comment, are you sure these should continue as `:commercial` (equivalent to a time trial) instead of `:gratis` (equivalent to freeware)?

Everything else looks great.

mbrgm:
@vitorgalvao I read the licensing agreement and it limits the use to one machine at a time. This is a limitation to the way how I'd interpret the `:gratis` stanza. So maybe fall back to `:closed`, or do you think `:gratis` is still a fit?

vitorgalvao:
> do you think `:gratis` is still a fit?

It is still free to use, so yes. `license` is more of a guideline to give a general view than anything else (for example, the `LGPL` is still included in the `GPL`), they’re not supposed to comprise such nuances.

vitorgalvao:
Everything looks great. Thank you.

PR #711, opened 2015-01-27
Upgrade IntelliJ Idea EAP to 14.0.3

zchee:


vitorgalvao:
You’re updating the wrong cask. It should (likely) be [this one](https://github.com/caskroom/homebrew-cask/blob/master/Casks/intellij-idea.rb).

PR #712, opened 2015-01-27
rstudio-preview: update to always pull latest

aguynamedryan:
Instead of submitting an updated formula for every update, I've
modified the formula to mimic the iterm-nightly formula.

I submitted a similar pull request (#688) that used my own URL
 but was told that was, understandably, an unsafe approach.

RStudio has kindly provided a link that will always point to the
latest Preview version of RStudio and I have updated the formula
to use this URL.

Thanks RStudio!

### DIFF
--- a/Casks/rstudio-preview.rb
+++ b/Casks/rstudio-preview.rb
@@ -1,8 +1,8 @@
 cask :v1 => 'rstudio-preview' do
-  version '0.99.174'
-  sha256 '50407f5f2fdcc84dab862d5c32e9264e8280338d625389ce9963e35cf83c7bc4'
+  version :latest
+  sha256 :no_check
 
-  url "https://s3.amazonaws.com/rstudio-dailybuilds/RStudio-#{version}.dmg"
+  url "http://www.rstudio.org/download/latest/preview/desktop/mac/RStudio-latest.dmg"
   homepage 'http://www.rstudio.com/ide/download/preview'
   license :unknown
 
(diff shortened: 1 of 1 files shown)


Casks/rstudio-preview.rb
```diff
@@ -1,8 +1,8 @@
 cask :v1 => 'rstudio-preview' do
-  version '0.99.174'
-  sha256 '50407f5f2fdcc84dab862d5c32e9264e8280338d625389ce9963e35cf83c7bc4'
+  version :latest
+  sha256 :no_check
 
-  url "https://s3.amazonaws.com/rstudio-dailybuilds/RStudio-#{version}.dmg"
+  url "http://www.rstudio.org/download/latest/preview/desktop/mac/RStudio-latest.dmg"
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed; please change these to single quotes.

aguynamedryan: I'm happy to make the change, in fact 2 weeks ago I would have used single quotes.

But now I'm getting confused.  Homebrew/homebrew requires double-quotes every where (at least, that's the request I got on my last pull request to them: https://github.com/Homebrew/homebrew/pull/35915#issuecomment-70172077)

Also, I recently read this article: http://viget.com/extend/just-use-double-quoted-ruby-strings

Please advise.

vitorgalvao: We have no relation to homebrew, organisationally. Different projects run by different people. Like I said, we reserve double quotes for cases where string interpolation is needed; makes it easier to make the distinction at a glance.

aguynamedryan:
Thanks!

PR #714, opened 2015-01-28
Updated earlybird sha256 to no_check

zerrot:
Earlybird nightly is updated daily, but the version number does not change. So the cask breaks if the sha check is performed.

### DIFF
--- a/Casks/earlybird.rb
+++ b/Casks/earlybird.rb
@@ -1,6 +1,6 @@
 cask :v1 => 'earlybird' do
   version '37.0a2'
-  sha256 'da335ece743f9211be7274e39a432cf4f1ec0886172cb0a731bd28489110964d'
+  sha256 :no_check  # required as upstream package is updated in-placea
 
   url "https://ftp.mozilla.org/pub/mozilla.org/thunderbird/nightly/latest-earlybird/thunderbird-#{version}.en-US.mac.dmg"
   name 'Earlybird'


Casks/earlybird.rb
```diff
@@ -1,6 +1,6 @@
 cask :v1 => 'earlybird' do
   version '37.0a2'
-  sha256 'da335ece743f9211be7274e39a432cf4f1ec0886172cb0a731bd28489110964d'
+  sha256 :no_check
```
vitorgalvao: Please add [a comment](https://github.com/caskroom/homebrew-cask/blob/6013ec0c29c59927d9ca50c29fbd410c9a23e970/Casks/adobe-air.rb#L3) here (done so people who learn by example know this should be done only in select cases).

zerrot: No problem. I'm getting ready to check in the updated commit. Take a look and let me know if I need to shorten the comments.

vitorgalvao: Make the comment exact, please, as we like to keep our comments consistent.

zerrot: Sorry I didn't see that. I will fix it and submit again.

zerrot: Okay should be good to go now.

vitorgalvao: It is. Thank you.

PR #719, opened 2015-01-30
Fix checksum in Google Japanese IME Dev build

zchee:


vitorgalvao:
Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/master/Casks/acorn.rb).

Would you care to correct the cask?

zchee:
Fix it !!

vitorgalvao:
Thank you.

PR #721, opened 2015-01-30
Sublime Text Dev 3067

yurikoles:
Additional question how to encode Sublime%20Text%20Build%203067.dmg from normal string? I ask this because for it is required to use custom url without #{version}


vitorgalvao:
I don’t understand what your question is. This cask could be better written as

```ruby
cask :v1 => 'sublime-text-dev' do
  version '3067'
  sha256 '19fea5cc5dad89a31f1b751bb7c358ffd7acdc4bc7b8c5b957631743a523a261'

  url "http://c758482.r82.cf2.rackcdn.com/Sublime%20Text%20Build%20#{version}.dmg"
```

I have no idea if that answers it or not.

ndr-qef:
Thanks, merged as 630cc21. There is no need to URL-encode the string — as @vitorgalvao mentioned, the existent `version` was poorly formulated.

PR #722, opened 2015-01-31
Updated various firefox casks

zerrot:
Updated firefox-beta and localized betas to version 36.0b5.
Removed firefox-ux because of broken download link and removal from Mozilla repository.


ndr-qef:
Thanks. For future reference, we generally prefer to isolate each update in their own commit: it is easier to address individual problems without blocking working changes, and works better in `git revert` and `git bisect`.

PR #736, opened 2015-02-05
Fix typo in cask

kstarsinic:
There was a typo in this cask, making it unusable.

$ brew cask audit
Error: Cask 'audio-hijack-pro' definition is invalid:
  invalid 'depends_on :macos' value: ":lion,"


vitorgalvao:
For future fixes, please always include the name of the cask in the commit’s short description.

PR #748, opened 2015-02-09
sublime-text3: Various improvements

jawshooah:


ndr-qef:
Very happy to see exhaustive zap stanzas. Just one thing: are you sure `conflicts_with` is necessary? From what I gather, different versions of Sublime Text should be able to coexist.

jawshooah:
Sublime Text 2 should be able to coexist with any iteration of 3, but I don't believe the same is true for multiple iterations of 3 (stable and dev).

They have the same app name, so installing one after the other would clobber the existing symlink. I'm also not sure that it's safe for stable and dev builds to share the same data directory, with caches and all. One would inevitably clobber some of the other's data, especially if they're running concurrently.

ndr-qef:
I am not very worried about app data, but you are right about symlink clobbering. @rolandwalker agrees that it is ground for `conflicts_with`.

PR #749, opened 2015-02-09
sublime-text-dev: Various improvements

jawshooah:
Add appcast, uninstall, zap, and conflicts_with stanzas. Add comment for
url indicating that download host is official vendor.


ndr-qef:
Thanks again. As in #748: is `conflicts_with` desirable here?

ndr-qef:
Thanks again.

PR #767, opened 2015-02-20
Update checksum and url for Sketch 3.3 Beta

arzatskis:


arzatskis:
Not sure what I've missed or done wrong following contribution guide.. Any tips?

ndr-qef:
Not to worry, your patch is fine. The problem is with our Travis setup.

PR #776, opened 2015-02-21
Added portugues Adobe Reader

alexandregz:


vitorgalvao:
These should all likely be `pt-br` or just `br`. `pt-pt` is different.

alexandregz:
sorry, Adobe only have pt_BR version

PR #791, opened 2015-02-28
Replace Homebrew constants with Hbc variables (fix test suite)

radeksimko:
This most likely caused tests in this tap to fail: 
https://github.com/caskroom/homebrew-cask/commit/b40b0c99e1ea43ca3a7126dd099356bf8e7c4720

so I'm fixing it in a very similar manner.


radeksimko:
Merged this myself, so we can carry on with other pull-requests that are currently all in red.

PR #830, opened 2015-03-08
Add TorBrowser.app v4.0.4 (Chinese)

Ngrd:


fanquake:
I'm curious, is there a project or something your working on that you need all the different Tor browser versions for?

Ngrd:
Not really. I only needed a few of those, but then decided why not add all of them if I already started. The Chinese version was the last.

PR #834, opened 2015-03-10
Upgrade arduino-beta to 1.6.1

bcomnes:
This formula is a little weird, because it installs the non-beta of arduino on snow-leapord or lower where you would expect to be getting a beta.  AFAICT the non java7 version is the same as the regular arduino formula.  This PR is just a version bump, but would you like me to remove that old code?


fanquake:
Version bump is fine for now. We can look at removing the old code in future.

PR #836, opened 2015-03-10
Fix bad header line

visualphoenix:
pokerstars-eu does not match file name - should be pokerstarseu


vitorgalvao:
Missed that. Thank you.

PR #838, opened 2015-03-11
[RFC] add dependency on homebrew docker formula for docker-machine?

visualphoenix:
Not sure if we want to add the explicit dependency on homebrew docker in case homebrew-versions users want to use a different version.

Thoughts?

### DIFF
--- a/Casks/dockermachine010.rb
+++ b/Casks/dockermachine010.rb
@@ -15,5 +15,6 @@ cask :v1 => 'dockermachine010' do
     system '/bin/chmod', '--', '0755', "#{staged_path}/docker-machine_darwin-amd64"
   end
 
+  depends_on :formula => 'docker'
   depends_on :arch => :x86_64
 end


vitorgalvao:
Is `docker` needed for this to work at all? If so, we don’t have much choice.

visualphoenix:
It is - but folks might want to pick specific docker versions. Not sure how
you handle that.

On Wednesday, March 11, 2015, Vítor Galvão <notifications@github.com> wrote:

> Is docker needed for this to work at all? If so, we don’t have much
> choice.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/caskroom/homebrew-versions/pull/838#issuecomment-78251935>
> .
>


-- 
Raymond "VisualPhoenix" Barbiero

vitorgalvao:
We don’t. If we went down that rabbit whole we’d need an absurd number of casks for each case. I’m not even sure this version freeze will be useful to anyone at all, in the first place. If a user wants a specific version of a dependency, they should install it themselves; we take care of doing stuff in the way that works best.

PR #843, opened 2015-03-12
Spotify 0.9.15.27

zzamboni:
This is the version of Spotify before they removed Applescript support from it.

### DIFF
--- a/Casks/spotify-old.rb
+++ b/Casks/spotify-old.rb
@@ -1,0 +1,20 @@
+cask :v1 => 'spotify-old' do
+  version '0.9.15.27'
+  sha256 '83a569463c7bafd2947471e0657f20f30f29eed2e7c69b31545c09999e8c1417'
+
+  url 'http://dl.filehorse.com/mac/mp3-and-audio/spotify/Spotify-0.9.15.27.dmg?st=chZgCLlSY_qP3kGNqQ6hEQ&e=1426196090&fn=Spotify.dmg'
+  name 'Spotify before they removed Applescript support'
+  homepage 'http://spotify.com/'
+  license :gratis
+
+  app 'Spotify.app'
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
The name shouldn’t be `-old`, but instead have the version ([example](https://github.com/caskroom/homebrew-versions/blob/master/Casks/virtualbox43691406.rb)).

Casks/spotify-old.rb
```diff
@@ -0,0 +1,20 @@
+cask :v1 => 'spotify-old' do
+  version '0.9.15.27'
+  sha256 '83a569463c7bafd2947471e0657f20f30f29eed2e7c69b31545c09999e8c1417'
+
+  url 'http://dl.filehorse.com/mac/mp3-and-audio/spotify/Spotify-0.9.15.27.dmg?st=chZgCLlSY_qP3kGNqQ6hEQ&e=1426196090&fn=Spotify.dmg'
+  name 'Spotify before they removed Applescript support'
```
vitorgalvao: This should be the name of the app (`Spotify`), not a place for comments.

Casks/spotify-old.rb
```diff
@@ -0,0 +1,20 @@
+cask :v1 => 'spotify-old' do
+  version '0.9.15.27'
+  sha256 '83a569463c7bafd2947471e0657f20f30f29eed2e7c69b31545c09999e8c1417'
+
+  url 'http://dl.filehorse.com/mac/mp3-and-audio/spotify/Spotify-0.9.15.27.dmg?st=chZgCLlSY_qP3kGNqQ6hEQ&e=1426196090&fn=Spotify.dmg'
```
vitorgalvao: Are you hosting this yourself? We can only take official links.

vitorgalvao:
Thank you, but we cannot take links from unofficial sources, here, as they pose a security risk.

PR #856, opened 2015-03-19
Add RubyMine EAP JDK bundled build

zchee:


fanquake:
Related #849

sebroeder:
PR #885 has a slightly more recent version and implements the changes proposed in #879.

radeksimko:
Thanks @zchee for the contribution, #885 had more recent version though, so I merged that one instead.

PR #877, opened 2015-03-25
Upgrade IntelliJ Idea CE with bundled JDK to 14.1

sebroeder:
- Change download URL to https

### DIFF
--- a/Casks/intellij-idea-ce-bundled-jdk.rb
+++ b/Casks/intellij-idea-ce-bundled-jdk.rb
@@ -1,17 +1,19 @@
 cask :v1 => 'intellij-idea-ce-bundled-jdk' do
-  version '14'
-  sha256 '09bb078252e2f6af6b58605ad3a380a71c8cc53f8e697e31fe03fcadb2152b07'
+  version '14.1'
+  sha256 '62c8b150ff82c31f2b6355a6f43b8ef69913b8f387cef6ddb4659622abb59a52'
 
-  url "http://download.jetbrains.com/idea/ideaIC-#{version}-jdk-bundled.dmg"
+  url "https://download.jetbrains.com/idea/ideaIC-#{version}-custom-jdk-bundled.dmg"
+  name 'IntelliJ IDEA'
   homepage 'https://www.jetbrains.com/idea/'
-  license :oss
+  license :apache
 
-  app "IntelliJ IDEA #{version} CE.app"
-  
+  app 'IntelliJ IDEA 14 CE.app'
+
   zap :delete => [
-                  "~/Library/Application Support/IdeaIC#{version}",
-                  "~/Library/Preferences/IdeaIC#{version}",
-                  "~/Library/Caches/IdeaIC#{version}",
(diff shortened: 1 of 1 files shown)


Casks/intellij-idea-ce-bundled-jdk.rb
```diff
@@ -1,17 +1,17 @@
...
-  app "IntelliJ IDEA #{version} CE.app"
-  
+  app "IntelliJ IDEA 14 CE.app"
+
   zap :delete => [
-                  "~/Library/Application Support/IdeaIC#{version}",
-                  "~/Library/Preferences/IdeaIC#{version}",
-                  "~/Library/Caches/IdeaIC#{version}",
-                  "~/Library/Logs/IdeaIC#{version}",
+                  "~/Library/Application Support/IdeaIC14",
```
vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

vitorgalvao: Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes. Better yet, these and the one in app should be changed to `#{version.to_i}`.

sebroeder: > Double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes. Better yet, these and the one in app should be changed to #{version.to_i}.

Done. Thanks for the hints.

sebroeder:
@vitorgalvao I am having second thought whether `#{version.to_i}` is actually better here than hard-coding the number `14`. I would like to keep the many intellij-based formulas as similar as possible and `#{version.to_i}` would not work for [intellij-idea-ce-eap.rb](https://github.com/caskroom/homebrew-versions/blob/0898d7a0b0b4724a1b3ab1af18d14134bf94a0a5/Casks/intellij-idea-ce-eap.rb).

The "official" homebrew-cask formula [intellij-idea.rb](https://github.com/caskroom/homebrew-cask/blob/0bf58228c27979189d8447d03be133729eabf14b/Casks/intellij-idea.rb) has the version hard-coded, too.

And in the RubyMine formula [rubymine.rb](https://github.com/caskroom/homebrew-cask/blob/0bf58228c27979189d8447d03be133729eabf14b/Casks/rubymine.rb) you can see that a) Jetbrains naming scheme is not very predictable (they use `70` instead of `7` for RubyMine) and b) errors are harder to find once the code gets to clever (the regex is wrong in the RubyMine formula because it evaluates to `704` instead of `70`.

So I am on the fence to reverting back to the hard-coded version number (and single quotes). What do you think? /cc @radeksimko

radeksimko:
> Jetbrains naming scheme is not very predictable

yeah, that sucks... +1 for hardcoding it then.

Generally speaking, if the vendor doesn't care about making these package versions/filenames predictable enough, we should not be spending time/effort by trying to work around that IMO.

vitorgalvao:
Agreed. No quarrels at all about hardcoding.

sebroeder:
Ready to be merged from my side.

Issue #879, opened 2015-03-26
Unify formulas for IntelliJ-based IDEs


sebroeder:
The following formulas are affected in homebrew-versions:
- [x] 0xdbe-eap (should be moved here from [homebrew-cask](https://github.com/caskroom/homebrew-cask/blob/8dba4ca011c49a471c47f364fe8712a64b962136/Casks/0xdbe-eap.rb)) #1207
- [x] appcode-eap-bundled-jdk #1070
- [x] appcode-eap #1071
- [x] clion-bundled-jdk #944, thanks @andystanton
- [x] clion-eap-bundled-jdk #1119, thanks @andystanton
- [x] clion-eap #1118, thanks @andystanton
- [x] intellij-idea-bundled-jdk #876
- [x] intellij-idea-ce-bundled-jdk #877
- [x] intellij-idea-ce-eap
- [x] intellij-idea-ce13
- [x] intellij-idea-eap #1073, thanks @narbsy
- [x] intellij-idea12
- [x] intellij-idea13
- [x] php-storm-bundled-jdk #1198 
- [x] phpstorm-eap #1088, thanks @bchatard
- [x] phpstorm7
- [x] pycharm-bundled-jdk #1017
- [x] pycharm-ce-bundled-jdk #1018
- [x] pycharm-ce-eap
- [x] pycharm-eap
- [x] rubymine-bundled-jdk #938
- [x] rubymine-eap  #886
- [x] rubymine-eap-bundled-jdk #885
- [x] rubymine633 (should this formula be renamed to rubymine6?) #887
- [x] webstorm-bundled-jdk #930, thanks @dnalot
- [x] webstorm-eap

Things that could/should be unified:

- Download URLs:
    - download over https instead of http (CloudFlare does not work over http, use direct download link starting with `https://download.jetbrains.com/`
- Add `name` stanzas
- License:
    - `:apache` for Community Editions (`-ce`)
    - `:commercial` for everything else 
- Remove plist hack, replace with the following caveat (see caskroom/homebrew-cask#9943 for details):
   
     ```ruby
    caveats <<-EOS.undent
      #{token} requires Java 6 like any other IntelliJ-based IDE.
      You can install it with

        brew cask install caskroom/homebrew-versions/java6

      The vendor (JetBrains) doesn't support newer versions of Java (yet)
      due to several critical issues, see details at
      https://intellij-support.jetbrains.com/entries/27854363
    EOS
    ```

[...]

PR #887, opened 2015-03-30
Clean up RubyMine 6.3.3 formula

sebroeder:
- Unify with other IntelliJ-based formulas (see #879)


dnalot:
I think you were right about renaming this.

PR #898, opened 2015-04-02
Upgrade all TorBrowser.app localizations to v4.0.6 and change signing key

Ngrd:
As was already mentioned in https://github.com/caskroom/homebrew-cask/pull/10200, a different signing key was used for this release including all localizations.


fanquake:
:+1:

PR #901, opened 2015-04-02
Upgrade IntelliJ Idea with bundled JDK to 14.1.1

tamird:
@radeksimko could you test to make sure I did the sha256 correctly?


radeksimko:
Pulling down your change... :horse_racing:

radeksimko:
LGTM

PR #916, opened 2015-04-09
Add cleanmymac version 2

toonetown:


radeksimko:
See #10465 - that's probably more complete as it solves the OSX version dependencies too?

radeksimko:
hm, wrong thread, sorry :)

radeksimko:
A few things to be sorted out before we merge this in:

 - where possible, we try to use versioned URLs instead of `latest` as it makes it easier for users to upgrade, which in this case makes
   - `version :latest` => `version '1.11'` (for classic) and `version '2.3.5'` (for 2)
   - `url` => `https://dl.devmate.com/com.macpaw.CleanMyMac/1.11/1417522595/CleanMyMacClassic-1.11.zip` and `https://dl.devmate.com/com.macpaw.CleanMyMac2/2.3.5/1427986644/CleanMyMac2-2.3.5.zip`
   - and appropriate `sha256` for both versions
 - [we do prefer HTTPS](https://github.com/caskroom/homebrew-cask/issues/10377) where available and most of the URLs (both `url` stanzas and `homepage`) seem to have HTTPS versions with no redirects
 - `appcast` stanzas will probably need to stay HTTP as these actually redirect back to HTTP when using HTTPS :disappointed:

vitorgalvao:
This looks simply like [a duplicate of the main cask](without extra options), so why add it here? The changes suggested by @radeksimko should be made to the main cask, though.

radeksimko:
@vitorgalvao Your link is pointing to nowhere, but I assume you meant to link to https://github.com/caskroom/homebrew-versions/blob/master/Casks/cleanmymac-beta.rb ?

In which case I agree we can leverage that cask, but considering the naming convention vendor seem to have, it should be renamed as well from `cleanmymac-beta` to `cleanmymac-2`.

vitorgalvao:
No, I meant [the cask in the official repo](https://github.com/caskroom/homebrew-cask/blob/master/Casks/cleanmymac.rb). That beta looks like it should be removed, though, it’s no longer a beta, and an old version, at that.

radeksimko:
In that case I disagree, we should keep version 2 in `homebrew-versions` in case someone deliberately wants to install that version instead of 3.

vitorgalvao:
That’s what I mean. In that case your suggested changes should also be implemented in this cask, but “classic” should be removed.

vitorgalvao:
@toonetown Do you plan on updating this?

toonetown:
I wasn't - but I can. The reason I submitted this pull frequent was just to keep track of the version that was in the main repo before updating it to version 3.

vitorgalvao:
I mean updating as in with the discussed changes.

toonetown:
I can do it tomorrow if you would like.

vitorgalvao:
Thank you. Please @mention me or @radeksimko (or both) when you do, so we get a warning to check back.

toonetown:
@vitorgalvao, @radeksimko - I updated with the suggestions given (use HTTPS, actual version, SHA sum, and remove the "classic" stuff)

vitorgalvao:
Thank you.

PR #920, opened 2015-04-13
Android Studio 1.2 Beta 3 added

falkorichter:


Casks/android-studio-canary.rb
```diff
@@ -1,8 +1,8 @@
 cask :v1 => 'android-studio-canary' do
-  version '1.0.0-rc4'
-  sha256 'eb945124105341729b4d4a4cec07332827f0ec601484e91593825db703b64635'
+  sha256 '0efab53a2b5510f9164987b363ec02a60382a55d3de9c2e0308d233e3ca26507'
 
-  url "https://dl.google.com/dl/android/studio/ide-zips/#{version}/android-studio-ide-135.1626825-mac.zip"
+  url "http://dl.google.com/dl/android/studio/ide-zips/1.2.0.8/android-studio-ide-141.1845774-mac.zip"
+
```
vitorgalvao: This empty line should go.

Casks/android-studio-canary.rb
```diff
@@ -1,8 +1,8 @@
 cask :v1 => 'android-studio-canary' do
-  version '1.0.0-rc4'
```
vitorgalvao: `version` should be updated. You’re removing it entirely.

falkorichter: but the link has totally changed. therefore I thought it would be cleaner to also remove the version and simply add the whole url as a string.

vitorgalvao: `version` can’t be removed. It’s an integral part of the cask, and every cask has one. Its goal is not to be interpolated, that’s a side-feature.

Casks/android-studio-canary.rb
radeksimko: Also, why are we changing the URL from HTTPS to HTTP? Can we keep it as HTTPS, please?

Casks/android-studio-canary.rb
```diff
@@ -1,18 +1,18 @@
...
-  app 'Android Studio.app'
+  app 'Android Studio.app', :target => 'Android Studio Canary.app'
 
   caveats <<-EOS.undent
     If you have Java 7 or above installed, you may want to use it as Android Studio JDK, for example:
 
     export STUDIO_JDK=/Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk
 
-    Please take a look at this post: http://tools.android.com/recent/androidstudio1rc3_releasecandidate3released
+    Please take a look at this post: http://tools.android.com/recent/androidstudio12beta3available
```
radeksimko: What's the point of updating the link here? The caveat is intentionally providing more details about Java issues & solutions. Look at the original link which has whole long paragraph about this.

The new link is not mentioning anything about that issue, therefore I suggest keeping it as is.

Casks/android-studio-canary.rb
```diff
@@ -1,18 +1,18 @@
...
+  sha256 '0efab53a2b5510f9164987b363ec02a60382a55d3de9c2e0308d233e3ca26507'
 
-  url "https://dl.google.com/dl/android/studio/ide-zips/#{version}/android-studio-ide-135.1626825-mac.zip"
+  url "http://dl.google.com/dl/android/studio/ide-zips/1.2.0.8/android-studio-ide-141.1845774-mac.zip"
+       
   homepage 'http://tools.android.com/download/studio'
   license :apache
 
-  app 'Android Studio.app'
+  app 'Android Studio.app', :target => 'Android Studio Canary.app'
```
radeksimko: If the naming convention from vendor says "Beta", can we use that instead of Canary, here? Or is there any particular reason for "canary"?

fanquake:
@falkorichter Could you make any further fixes and please squash all your commits together.

vitorgalvao:
Done in another PR.

PR #922, opened 2015-04-14
Add unity4 and unity45 versions.

mrmachine:
Now that Unity 5 is out, I thought a version for Unity 4.x would be good. Also a version for Unity 4.5.x because 4.6 introduced some backwards incompatible changes and others might also have projects that haven't yet been updated.

### DIFF
--- a/Casks/unity4.rb
+++ b/Casks/unity4.rb
@@ -1,0 +1,12 @@
+cask :v1 => 'unity4' do
+  version '4.6.4'
+  sha256 'd5d840f30d0987b3aef29dc3b651141cb5fb77fc3c28405b5ff667e03b01360a'
+
+  url "http://download.unity3d.com/download_unity/unity-#{version}.dmg"
+  homepage 'http://unity3d.com/unity/'
+  license :commercial
+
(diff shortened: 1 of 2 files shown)


Casks/unity4.rb
```diff
@@ -0,0 +1,12 @@
+cask :v1 => 'unity4' do
```
radeksimko: Shouldn't this be called `unity46` if the version is `4.6.4` and the other cask with version `4.5.5` is called `unity45`?

vitorgalvao: I noticed that but reasoned it wouldn’t be harmful this way because it means “the last of version 4”, which means you care less about specificity. With version `4.5` you care more about the minor version increments.

radeksimko: I see, that makes sense then. :+1:

mrmachine: Yep. That's the reason. `unity4` is "give me any version of Unity 4.x (anything earlier than Unity 5). `unity45` is "give me any version of Unity 4.5.x (anything earlier than Unity 4.6).

PR #925, opened 2015-04-14
Added Tableau 8

mattnelson:


Casks/tableau8.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'tableau8' do
+  version '8.2.9'
+  sha256 '2644c7b9fd6c453d00f6641ddb681b87ba5cdfa16c6d840163295fc176f54ee1'
+
+  url 'https://downloads.tableau.com/esdalt/8.2.9/TableauDesktop.dmg'
```
fanquake: You can use #{version} in the URL here

Casks/tableau8.rb
```diff
@@ -0,0 +1,11 @@
...
+  name 'Tableau'
+  homepage 'http://www.tableausoftware.com/'
```
fanquake: It looks like there's a https link available, can you use that instead.

mattnelson:
Addressed comments in 7b14e98209a198f724aab61e75fe6fae435f4b69

Also created separate casks for the 8.2 and 8.3 streams.

mattnelson:
Anything else holding up this pull?

nanoxd:
Sorry about that. Looks good!

PR #944, opened 2015-04-26
Add CLion 1.0 with bundled JDK 1.8

andystanton:


sebroeder:
Awesome work, thanks. Looks good to me and also works as expected on my machine. The only teeny-tiny nitpick I have is that the homepage url could also be https, but that’s a nice to have, not a must. :+1:

andystanton:
Not a problem @sebroeder. I've changed the homepage url to ```https``` and fixed up the commit.

PR #953, opened 2015-05-02
Add Omnipresence beta track (currently 1.4)

mauricerkelly:
This creates a new cask for the Omnipresence beta (at v1.4 at the
time of creation).

### DIFF
--- a/Casks/omnipresence-beta.rb
+++ b/Casks/omnipresence-beta.rb
@@ -1,0 +1,13 @@
+cask :v1 => 'omnipresence-beta' do
+  version '1.4.x-r232733'
+  sha256 'c21289a7a90b4efc8f0eb9cf3775bb4bb19145688229cd886718881b2c34377a'
+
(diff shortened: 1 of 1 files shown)


Casks/omnipresence-beta.rb
```diff
@@ -0,0 +1,14 @@
+cask :v1 => 'omnipresence-beta' do
+  version '1.4.x'
+  release_num = 'r232733'
```
vitorgalvao: This should go, and `version` should just be `1.4.x-r232733`. There’s no advantage in separating them.

mauricerkelly: @vitorgalvao No problem - I wasn’t sure if there were negative effects from putting too much in the version field (for example, the install directory is changed), but I’ve updated with your guidance. Much cleaner now :+1:

vitorgalvao: Thank you.

PR #981, opened 2015-05-13
Fixing version, sha256, and url for libreoffice43 cask

ianwalter:
This cask initially mirrored the cask in homebrew-cask but it turns out version 4.3.6 is no longer hosted, it should actually be 4.3.7. I've updated this cask and I will create a separate pull request to update homebrew-cask.


vitorgalvao:
What’s the value of having this, then?

ianwalter:
homebrew-cask installs the 32-bit version of libreoffice 4.3.7 only if the user is installing on a 32-bit version of OS X, otherwise it installs 4.4.x. This cask allows the user to explicitly install libreoffice 4.3.7 (my company needs this version because of a bug in 4.4.x). Thanks.

PR #1005, opened 2015-05-22
Add Tower beta channel (currently v2.2.2)

mauricerkelly:
This commit adds the Tower beta channel to Versions. It is based on
the main `tower.rb` in `homebrew-cask`, with updates for the beta
appcast.

### DIFF
--- a/Casks/tower-beta.rb
+++ b/Casks/tower-beta.rb
@@ -1,0 +1,24 @@
+cask :v1 => 'tower-beta' do
+  version '2.2.2-281'
+  sha256 '83fdbb887c394ed0f76bf91c66972d28bc6d36f144c21dfd582ccb09d5e19864'
+
+  # amazonaws.com is the official download host per the vendor homepage
+  url "https://fournova-app-updates.s3.amazonaws.com/apps/tower2-mac/281-3f2b7672/Tower-2-#{version}.zip"
+  appcast 'https://updates.fournova.com/updates/tower2-mac/beta'
+  name 'Tower'
+  homepage 'http://www.git-tower.com/'
+  license :commercial
+
+  app 'Tower.app'
+  binary 'Tower.app/Contents/MacOS/gittower'
+
+  zap :delete => [
+                  '~/Library/Application Support/com.fournova.Tower2',
+                  '~/Library/Caches/com.fournova.Tower2',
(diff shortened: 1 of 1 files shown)


Casks/tower-beta.rb
```diff
@@ -0,0 +1,23 @@
+cask :v1 => 'tower-beta' do
+  version '2.2.2-281'
+  sha256 '83fdbb887c394ed0f76bf91c66972d28bc6d36f144c21dfd582ccb09d5e19864'
+
+  url "https://fournova-app-updates.s3.amazonaws.com/apps/tower2-mac/281-3f2b7672/Tower-2-#{version}.zip"
```
vitorgalvao: Needs a `# amazonaws.com is the official download host per the vendor homepage` comment above this line ([example](https://github.com/caskroom/homebrew-cask/blob/cb487ad405f74cb1f26b93e3f3e8412c82a22e34/Casks/0ad.rb#L5)).

Where did you find the link, though?

mauricerkelly: It's embedded in the appcast link. I have a beta version installed, and I knew that there were Sparkle updates for it. I saw there was an appcast link for `stable`, so I guessed at `beta`.

Will add the requested comment, and also update the main `tower.rb` as I created this cask based on it, so it'll be missing it as well.

mauricerkelly: @vitorgalvao Updated the PR. Are you wary about the source of the URL?

vitorgalvao: I’m not, because I trust you, but users should never have to trust us — everything should be verifiable independently by them, and we make a point to make it so.

mauricerkelly: Do you want me to just make it a `:latest` instead?




BTW, I often grab the download URLs from the safari-downloaded package. The quarantine metadata usually illustrates the real URL instead of an anonymous latest URL.

vitorgalvao:
It was easy to check from the appcast, so lets incude it.

mauricerkelly:
@vitorgalvao I'm guessing that the idea here is that when the URL isn't obviously from the vendor, we want the Cask user to trust that it's not some malware URL that has been hosted from an unscrupulous person?

vitorgalvao:
Precisely. That is why the comment must be added *and* the maintainer that merges it must always verify it.

PR #1029, opened 2015-05-30
Update Dash beta to 2.3.0

mauricerkelly:
This commit updates the version, license and sha256 stanzas, and also adds
a note to explain the source of the download URL.

### DIFF
--- a/Casks/dash-beta.rb
+++ b/Casks/dash-beta.rb
@@ -1,10 +1,12 @@
 cask :v1 => 'dash-beta' do
-  version '2.2.1'
-  sha256 '2a652e850c1d9bd99977b6bc2b842ff7eeb07ba401a1bceab43d6998a60f6756'
+  version '2.3.0'
+  sha256 '5434eb1eb8a1429d4e1e6bcdb5b76a7b5930ceab1bbd94f41cb6a41e454e7512'
 
-  url 'https://dl0tgz6ee3upo.cloudfront.net/production/app/builds/002/376/285/original/e1c44c93cf9c3d49ff42b6349e56630c/Dash_Beta.app.zip'
+  # cloudfront.net is the official download host per the Dash beta page:
+  # https://rink.hockeyapp.net/download/ad59cd8c60434509b4b58951e1b7b483
+  url 'https://dl0tgz6ee3upo.cloudfront.net/production/app/builds/004/633/802/original/537af3eefba4513c44aedbab7606c3bf/Dash_Beta.app.zip'
   homepage 'http://kapeli.com/dash'
-  license :unknown
+  license :commercial
 
   app 'Dash Beta.app'
 end


Casks/dash-beta.rb
```diff
@@ -1,10 +1,12 @@
 cask :v1 => 'dash-beta' do
-  version '2.2.1'
-  sha256 '2a652e850c1d9bd99977b6bc2b842ff7eeb07ba401a1bceab43d6998a60f6756'
+  version '2.3.0'
+  sha256 '5434eb1eb8a1429d4e1e6bcdb5b76a7b5930ceab1bbd94f41cb6a41e454e7512'
 
-  url 'https://dl0tgz6ee3upo.cloudfront.net/production/app/builds/002/376/285/original/e1c44c93cf9c3d49ff42b6349e56630c/Dash_Beta.app.zip'
+  # cloudfront.net is the official download host per the Dash beta page:
+  # https://rink.hockeyapp.net/download/ad59cd8c60434509b4b58951e1b7b483
```
vitorgalvao: Why is this useful? It leads to a login page.

Casks/dash-beta.rb
```diff
@@ -1,10 +1,12 @@
 cask :v1 => 'dash-beta' do
-  version '2.2.1'
-  sha256 '2a652e850c1d9bd99977b6bc2b842ff7eeb07ba401a1bceab43d6998a60f6756'
+  version '2.3.0'
+  sha256 '5434eb1eb8a1429d4e1e6bcdb5b76a7b5930ceab1bbd94f41cb6a41e454e7512'
 
-  url 'https://dl0tgz6ee3upo.cloudfront.net/production/app/builds/002/376/285/original/e1c44c93cf9c3d49ff42b6349e56630c/Dash_Beta.app.zip'
+  # cloudfront.net is the official download host per the Dash beta page:
```
vitorgalvao: Please change the comment to be exactly `# cloudfront.net is the official download host per the vendor homepage`, as it is useful to have them all be the same.

Casks/dash-beta.rb
mauricerkelly: If you're signed in to HockeyApp and are part of the Dash beta you would have the download link. I did forget about the fact that it wouldn't be relevant to people who are not part of the Dash beta, but I guess in that case they're possibly not supposed to be downloading the software.

Casks/dash-beta.rb
mauricerkelly: Sure. My reasoning here was simply that the vendor homepage does not have the link, only the vendor's beta program through HockeyApp.

Casks/dash-beta.rb
vitorgalvao: Ah, in that case the cask must be deleted, it should never have got through, since the URL can’t be independently confirmed. It should be deleted from here and added to [caskroom/unofficial](https://github.com/caskroom/homebrew-unofficial) instead, as [per the tules](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#unofficial-vendorless-or-walled-builds). Would you care to do it?

mauricerkelly: Sure. I didn't realise that since I was just updating. Note that the actual download link is publicly accessible, but the beta announcements aren't necessarily so. It's always seemed like security through obscurity to me!

vitorgalvao: Sure, the links must always be publicly accessible for us to get them, but the problem is when users can’t confirm them, as we talked previously.

Exceptions made when the download url is on the same host as the homepage, in which case you can trust it even if behind a registration wall.

mauricerkelly: Cool, cool. Don't get me wrong, I'm not trying to dispute or anything. I'm only wary that by raising this, I've "spoiled" this cask for some people, so I want to make sure there's no reason it should stay here before I move it away :grinning:

vitorgalvao: Oh, I know, I’ve merged enough from you to know you understand our reasoning and that if you disagree, it would be for good reason. There’s no harm in moving it: it’s just a different tap under the same organization, after all.

mauricerkelly:
Closing this PR in favour of #1030

PR #1030, opened 2015-05-30
Remove Dash Beta for moving to Unofficial

mauricerkelly:
While making an update to `dash-beta.rb`, @vitorgalvao pointed out
that the private nature of the Dash beta meant that it belonged in
caskroom/homebrew-unofficial instead.

This commit removes the `dash-beta.rb` file pending its move.


mauricerkelly:
The PR for addition to unofficial is caskroom/homebrew-unofficial#45

vitorgalvao:
Thank you.

PR #1031, opened 2015-05-31
android-studio-canary 1.3.0.0 fixed sha256 mismatch

melnikovdv:
fixed sha256 mismatch


vitorgalvao:
Please [amend](http://stackoverflow.com/a/179147/1661012) the commit message to include the name of the updated cask.

PR #1037, opened 2015-06-03
Added Adobe CS6 Collection Editions & Language Versions

winkelsdorf:
Here they are. Sadly much work, as I am still waiting for any kind of response to https://github.com/caskroom/homebrew-cask/issues/11366. Would have been only 45% of the work now needed.

Manually tested many of them in fresh VMs, but not all due to the huge amount of download volume and time needed. But I ensured they are logically correct. That said, I tested all of the English and German versions and some of the Spanish and French versions.


vitorgalvao:
What a ton a work, here. Thank you.

winkelsdorf:
Indeed 😂 You're welcome! Glad I could help, hopefully some Users do find the Casks useful, too :)

PR #1051, opened 2015-06-09
don't try to install java6 on 10.11

yurikoles:
fix https://github.com/caskroom/homebrew-cask/issues/11669


vitorgalvao:
Thank you.

PR #1117, opened 2015-07-09
Add cask for IntelliJ IDEA 14.0 (old version).

adjohnson916:
Some users might prefer to use 14.0.4, the latest 14.0.x release, rather than any of the latest 14.1.x releases.


vitorgalvao:
To include this, these need to follow [these rules](https://github.com/caskroom/homebrew-versions/issues/879).

However, why would someone prefer this version? Unless there are crippling bugs or some huge reason to keep this (and you even say “some users might prefer”), I see no great value in adding this, but a great disadvantage in overhead. Let’s not include every version possible.

Thank you for the contribution, sill.

adjohnson916:
@vitorgalvao This is for the Ultimate version which is commercial software where purchased licenses are valid only up to the latest minor release before expiration, then no longer work with latest versions. This 14.0 is only one minor version behind the latest 14.1 release. So this would probably still be useful to many users - perhaps more so than the much older 12 and 13 versions that already have casks in this repository.

* https://sales.jetbrains.com/hc/en-gb
* http://blog.jetbrains.com/idea/2013/12/intellij-idea-personal-licensing-changes/

If you are more open to it after reading this, I could work on making changes to follow [those rules](https://github.com/caskroom/homebrew-versions/issues/879) you cited.

vitorgalvao:
Sounds fair. Please make the changes, then.

fanquake:
@adjohnson916 Are you planning on making the necessary changes?

adjohnson916:
@fanquake Not currently. Time consuming.

fanquake:
Closing this as changes need to be made, and the branch needs a rebase.

PR #1120, opened 2015-07-09
fix clion-bundled-jdk sha & update conflicts_with.

andystanton:


vitorgalvao:
Pinging @sebroeder.

sebroeder:
@andystanton good catch regarding the SHA1 mismatch, I could reproduce this on my system. Could you please revert the change go the `conflicts_with` stanza please? This cask only conflicts with `clion` since you can install CLion and CLion EAP next to each other.

@vitorgalvao Doesn't the travis build check for SHA1 mismatches?

vitorgalvao:
`sha256` you mean, but no, it doesn’t. That’d require downloading every file, and that’d take a lot of bandwidth. Travis checks for correctness in the cask itself, not in the results.

jawshooah:
It's for Travis to determine which Casks were added or changed by a PR and
verify only those hashes.

vitorgalvao:
I didn’t mean every time, if that’s what you mean, but that’s still a lot of bandwidth. I haven’t used Travis much so I don’t have a strong opinion either way, but even if we could do it for free I'm not sure I’d be comfortable with it (maybe we’re even supposed to, and that’s OK, but for such a trivial matter, not sure it’s worth it).

jawshooah:
In any case, this is way off-topic. I'll open a thread on the main repo for further discussion.

andystanton:
I'm closing this PR as 1.0.5 is out anyway and the ```conflicts_with``` stanza change was incorrect.

A fresh PR for 1.0.5 will be a bit cleaner.

PR #1157, opened 2015-07-21
Update Torbrowser alpha and add some caveats

olbrew:
This updates TorBrowser to the latest alpha and renames it to that from experimental as that is the name used by the TorProject itself.
It also changes the language to en-US instead of nl and adds some caveats to update with built-in update mechanism thereafter to keep local settings.


Casks/torbrowser-alpha.rb
```diff
@@ -0,0 +1,17 @@
...
+  version '5.0a3'
+  sha256 '79f5b53f51124d483fd19e33eca1fdbc8fc37a8e2e0d4b55844fd45e8f954ded'
+
+  url "https://dist.torproject.org/torbrowser/#{version}/TorBrowser-#{version}-osx64_en-US.dmg"
+  gpg "#{url}.asc",
+      :key_id => 'ef6e286dda85ea2a4ba7de684e2c6e8793298290'
+  name 'Tor Browser'
+  homepage 'https://www.torproject.org/projects/torbrowser.html'
+  license :oss
+  app 'TorBrowser.app'
```
vitorgalvao: There should be an empty line before `app`.

fanquake:
@olbrew Can you please squash your commits

olbrew:
Done!

PR #1160, opened 2015-07-21
Update kicad-nightly.rb

abardagjy:
Updated kicad-nightly to r5967.20150721-035418.dmg (current). Also because previous build was DOA (error https://i.imgur.com/VAiOIFw.png).


abardagjy:
Thank you!

PR #1168, opened 2015-07-28
Use the new package for java 6 that works on el capitan

toonetown:
This build will run and install even if rootless is enabled on the drive.

### DIFF
--- a/Casks/java6.rb
+++ b/Casks/java6.rb
@@ -1,14 +1,12 @@
 cask :v1 => 'java6' do
   version '1.6.0_65'
-  sha256 '97bc9b3c47af1f303710c8b15f2bcaedd6b40963c711a18da8eac1e49690a8a0'
+  sha256 '1b7b88c7c7ca3a1c50eacee1977dee974a50157ff7c88d0e73902bd98fc58a86'
 
-  url 'http://support.apple.com/downloads/DL1572/en_US/JavaForOSX2014-001.dmg'
-  homepage 'http://support.apple.com/kb/DL1572'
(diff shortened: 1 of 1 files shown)


Casks/java6.rb
```diff
@@ -1,14 +1,12 @@
 cask :v1 => 'java6' do
   version '1.6.0_65'
-  sha256 '97bc9b3c47af1f303710c8b15f2bcaedd6b40963c711a18da8eac1e49690a8a0'
+  sha256 '1b7b88c7c7ca3a1c50eacee1977dee974a50157ff7c88d0e73902bd98fc58a86'
 
-  url 'http://support.apple.com/downloads/DL1572/en_US/JavaForOSX2014-001.dmg'
-  homepage 'http://support.apple.com/kb/DL1572'
+  url 'https://support.apple.com/downloads/DL1572/en_US/javaforosx.dmg'
+  homepage 'https://support.apple.com/kb/DL1572?locale=en_US'
```
vitorgalvao: `?locale=en_US` should be removed. It serves no purpose.

ming13:
@vitorgalvao, can you please take a look again? Seems like the previous package location is down and not working at all.

Ngrd:
I hope that @vitorgalvao won't mind, I will merge this for now. The previous URL is not available anymore and this cask was tested and works very well. Thank you for your contribution!

vitorgalvao:
> I hope that @vitorgalvao won't mind, I will merge this for now.

Not at all, @Ngrd, I appreciate it. When in doubt, feel free to merge first (if you think the cask is correct) and ask later (if even so you’d like something clarified). I trust your judgement.

PR #1174, opened 2015-07-30
Update AppCode EAP to build 141.2000

mauricerkelly:
The previous EAP build has expired so this commit updates the version
to include the latest build, as well as the sha256 stanza.


jawshooah:
Thank you for the contribution. It was merged directly as 78346b5a202cdefeb12e2afb9772f62fc1248b40 to keep commit history cleaner. Your contribution is still credited to you.

PR #1186, opened 2015-08-05
Add kaleidoscope1 cask

samueldodson:


Casks/kaleidoscope1.rb
```diff
@@ -0,0 +1,10 @@
+cask :v1 => 'kaleidoscope1' do
+  version '1.1.8'
+  sha256 '105c0c05062536fb8c5267a61ed7155d8273e29e650fd655084612acc326babd'
+
+  url "http://d23o3xv0lvihc9.cloudfront.net/KSUpdates/1.1.8_57/Kaleidoscope.zip"
```
vitorgalvao: Needs a `# cloudfront.net is the official download host per the vendor homepage` comment above this line ([example](https://github.com/caskroom/homebrew-cask/blob/cb487ad405f74cb1f26b93e3f3e8412c82a22e34/Casks/0ad.rb#L5)).

Also, double quotes are reserved for cases where string interpolation is needed. Please change these to single quotes.

vitorgalvao:
Please point to where the download URL is located for this version, so we can verufy it.

samueldodson:
The download URL is http://www.kaleidoscopeapp.com/download_ks1, which is forwarded to http://d23o3xv0lvihc9.cloudfront.net/KSUpdates/1.1.8_57/Kaleidoscope.zip.

Ngrd:
Thank you for pointing out the location of the download URL.

samueldodson:
Thanks, Victor. I've added the closing quotation mark to the url stanza.

Ngrd:
Thank you for the contribution and all the patience in making all those changes. You are awesome!

PR #1190, opened 2015-08-09
Add dash 2 cask

nicday:


Casks/dash-2.rb
```diff
@@ -0,0 +1,21 @@
+cask :v1 => 'dash-2' do
```
vitorgalvao: No `-` before numbers.

nicday: No worries, `-` removed

Casks/dash2.rb
```diff
@@ -0,0 +1,21 @@
+cask :v1 => 'dash2' do
+  version '2.2.5'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

fanquake:
@nicday Could you make the required changes and squash your commits?

PR #1199, opened 2015-08-13
Change Arduino Beta cask to Arduino Nightly

fanquake:
Arduino has moved away from Beta versions, and is now doing nightly (hourly) builds.


Casks/arduino-nightly.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'arduino-nightly' do
+  version :latest
+  sha256 :no_check
+
+  url "http://downloads.arduino.cc/arduino-nightly-macosx.zip"
```
vitorgalvao: These need to be single quotes.

PR #1208, opened 2015-08-18
added virtualbox 5.0.0 (old version)

sunny4280:
- add virtualbox and virtualbox-extension-pack (version 5.0.0-101573)


Ngrd:
Thank you for your contribution. However, It's is much better to use separate commits for each cask being modified or added even though those are dependent with each other. This will make the commit history much cleaner and also easier to search through changes being made.

PR #1234, opened 2015-08-31
Remove 1Password 3

fanquake:
1Password 3 is available from the main repository.

https://github.com/caskroom/homebrew-cask/blob/master/Casks/1password.rb


vitorgalvao:
[Same](https://github.com/caskroom/homebrew-versions/pull/1235#issuecomment-136321503).

PR #1235, opened 2015-08-31
Remove Cleanmymac2

fanquake:
Cleanmymac2 is available in the main repo.
https://github.com/caskroom/homebrew-cask/blob/master/Casks/cleanmymac.rb


vitorgalvao:
It is available there, yes, but [this is a special case](https://github.com/caskroom/homebrew-versions/pull/916). The version 2 in the main repo will only be installed in certain OS X versions. Being a commercial app, this is here for people who have newer OS X versions but a license only for the old CleanMyMac version.

PR #1236, opened 2015-08-31
Fix missing quote in adium-nightly16

toonetown:


Ngrd:
Thank you for noticing and fixing that.

PR #1239, opened 2015-09-02
Add Umsatz (Pro, Mini) Latest

lukasbestle:
There is also a Standard version which I have PRed to the main repo.

### DIFF
--- a/Casks/umsatz-mini.rb
+++ b/Casks/umsatz-mini.rb
@@ -1,0 +1,11 @@
+cask :v1 => 'umsatz-mini' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://umsatz-programm.de/ladungen/UmsatzMini2015.zip'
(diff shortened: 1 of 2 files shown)


Casks/umsatz-mini.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'umsatz-mini' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://umsatz-programm.de/ladungen/UmsatzMini2015.zip'
+  name 'Umsatz Mini 2015'
+  homepage 'http://umsatz-programm.de'
+  license :freemium
```
vitorgalvao: Is this a free app that works indefinitely but with limitations that can be removed by paying?

lukasbestle: Yes, it is.

PR #1240, opened 2015-09-02
Add Reeder Beta

cprecioso:


vitorgalvao:
Lets leave this one out. We had reeder at other points in the past, but having it means having it until it is released, and then removing it again indefinitely, which isn’t that great of an experience.

On another note, this one isn’t `:commercial`, but `:gratis`. You cannot pay to remove its limitations. What matters is the license of the binary itself, and for all intents and purposes, you cannot buy the beta.

cprecioso:
Ok, added it to my own tap at cprecioso/homebrew-custom

PR #1243, opened 2015-09-03
Update libreoffice43.rb

pquentin:
Older versions are now on https://downloadarchive.documentfoundation.org/. The sha256 did not change, so it's the exact same version.


pquentin:
Thank you for the merge @vitorgalvao.

PR #1255, opened 2015-09-10
Update appcode-eap-bundled-jdk to v142.4675.7

lumaxis:


fanquake:
This needs rebasing.

fanquake:
Thankyou for the contribution, however I've now removed the eap-bundled-jdk cask, as the eap is now only available bundled.

PR #1300, opened 2015-10-02
Delete TextExpander3

fanquake:
Available from https://github.com/caskroom/homebrew-cask/blob/master/Casks/textexpander.rb


vitorgalvao:
This one should stay, unfortunately, similarly to other commercial apps in this situation. Version 3 is available in the main repo yes, **but** it will only install if you’re using Snow Leopard. However, if you’re using a later OS X version but still want to use your version 3 license, you need it as a separate cask.

fanquake:
Right, I keep forgetting that use case..

joech4n:
Did this actually get removed? I can't seem to be able to find it in https://github.com/caskroom/homebrew-versions/tree/master/Casks ...

PR #1346, opened 2015-10-16
Add firefox version 4

adius:


vitorgalvao:
Why? What value does it add to add versions `4` and `5`?

They don’t even have their own homepages here, and they should.

vitorgalvao:
Thank you for going through the trouble of making these, but we’ll refuse them. I see absolutely no use case for having them, and they’re just making the repo even more unmanageable.

They don’t even have their own homepages (because they’re not sufficiently relevant), and since firefox auto-updates, there isn’t even a point.

adius:
The point is to test browser-compatibility of your websites.
I thought this was the whole point of this repo to contain old versions of software for testing purposes.

vitorgalvao:
Thank you for going through the trouble of making these, but we’ll refuse them. I see absolutely no use case for having them, and they’re just making the repo even more unmanageable. Just imagine if after that someone started adding a cask for every language of every version!

They don’t even have their own homepages (because they’re not sufficiently relevant), and since firefox auto-updates, there isn’t even a point.

All that said, if you see a value in having so many old versions of Firefox, you can make a separate tap, for them. If you explain you reasoning for wanting so many of them available in this manner, we might even publicise said tap in the repo.

adius:
…I'd rather refuse the language versions. They are much less helpful in my opinion. ^^

vitorgalvao:
> I thought this was the whole point of this repo to contain old versions of software for testing purposes.

No, that is not correct at all. Were that the case, we’d call this `caskroom/homebrew-testing`, or the like. The point of this repo is to provide alternate versions to the casks already in the main repo, that’s it, testing does not enter the conversation.

Some types of casks we have here:

Alternative languages, old versions to solve to compatibility issues and old versions of commercial software (in case you don’t want to buy a license for the newest one) are just some of cases of casks accepted here.

> …I'd rather refuse the language versions. They are much less helpful in my opinion. ^^

So you think there are more people wanting old versions of a web browser than people wanting said browser in their language? I seriously doubt that. Actually, we *know* that to not be the case. Case in point: alternative languages have been submitted for a long while, but extremely old versions haven’t.

> The point is to test browser-compatibility of your websites.

Again (though I have edited the post at the same time you posted that), Firefox auto-updates, that isn’t a valid concern.

adius:
http://gs.statcounter.com/#desktop-browser_version-ww-monthly-201510-201510-bar
Well, to me it looks like around a third of the people do not browse with an up to date version.
…but hey, it's obviously not a valid concern.
Anyway, I'll make a new repository with old versions of all browsers.
If you change your mind one day, feel free to merge.

vitorgalvao:
> http://gs.statcounter.com/#desktop-browser_version-ww-monthly-201510-201510-bar
> Well, to me it looks like around a third of the people do not browse with an up to date version.

We’re specifically discussing Firefox, here. On that list, I see two Firefox versions, and none of them under 40.

> …but hey, it's obviously not a valid concern.

So no, adding the 39 (or whatever) versions before that is not a valid concern.

> If you change your mind one day, feel free to merge.

Unlikely. Like I said, this repo is already near unmanageable, and a big chunk of that problem are mozilla apps (alternative languages, nightlies). Translation languages are finite, version numbers are not. Your list of submissions would only grow, turning this effectively into a Firefox tap. Also, again, there simply isn’t a demand for this. Don’t forget that after you submit and we merge, *we* have to maintain them, and having an ever growing list of older versions of a self-updating web browser is a very poor use of resources for an already small team.

PR #1374, opened 2015-10-25
Syntax error fix (causes `brew cask audit` hard failure)

Ocramius:


vitorgalvao:
Please [amend](http://stackoverflow.com/a/179147/1661012) the commit message to include the name of the updated cask. Please @mention me after you do, so I get an email to check back.

Ocramius:
Why do I need to mention the affected cask in the commit message?
There is no change specific to URL or data of the cask information itself...

vitorgalvao:
Because without it it seems like you’re doing a general fix to the core, while it is not. A single cask is causing the issue, and we should know that at a glance.

Ocramius:
Since I'm doing this from the web editor, I opened #1375 to have the correct commit message (instead of amending)

vitorgalvao:
Merged this one by mistake (saw the reference and confused it for the fixed commit). No big deal. It would be better to have the cask specified, but it isn’t breaking.

Thank you.

Ocramius:
Alright, thanks for clarifying btw :-)

PR #1379, opened 2015-10-27
Update royal-tsx-beta.rb

Asshad:
Hash update

### DIFF
--- a/Casks/royal-tsx-beta.rb
+++ b/Casks/royal-tsx-beta.rb
@@ -1,6 +1,6 @@
 cask :v1 => 'royal-tsx-beta' do
   version '2.2.0.1000'
-  sha256 '4f46388d574cc4b4bcc8881b5fdb086d0c8ea47a6cc5563ebe6cc4a84795d363'
+  sha256 'e547054b6ce818d1429701804069d0aec7c5e494f9b757e2068093bb6ca4c9a2'
 
   url "http://v2.royaltsx.com/updates/royaltsx_#{version}.dmg"
   name 'Royal TSX'


Asshad:
Hi Vitor,

Thanks for the update. But it didn't work as freemium. Hmm.

adityadalal924:
Seems Freemium to me
screen shot 2015-10-27 at 11 14 19 am

vitorgalvao:
It isn’t `:freemium`, no, it is actually `:gratis`.

[The business model of the company is irrelevant, what matters is the app itself](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#license-stanza-details).

Asshad:
Hi Vitor,

Could you please enlighten me with those? 
"It isn’t :freemium, no, it is actually :gratis."

I managed to pull and installed but still,:freemium,it's limited to 10 connection only.

Any other way to get that as :gratis ?

vitorgalvao:
> Any other way to get that as :gratis ?

It doesn’t matter. That’s just a reference in the cask, it has no bearing in your use of the app.

vitorgalvao:
Looks like it is `:freemium` indeed. I see now what I missed it that text of theirs. Thought the app without limitation would be a different binary.

Asshad:
Yeah, I guess the app has been patched. That'll only work for beta checksum. 

If you have any workaround, it'd be lovely if you could share with me.

vitorgalvao:
Workaround for what?

Asshad:
The one without limitation(Royal Tsx)

adityadalal924:
You can buy Royal TSX [here](http://www.royalapplications.com/ts/osx/buy)

The `license` stanza refers to the app; `freemium` means that the app has some free features, and you need to pay for the rest. See [here](https://en.wikipedia.org/wiki/Freemium) if you are still confused about this type of business model.

Asshad:
Alritey. Got it. Thanks sir.

PR #1385, opened 2015-10-30
webkit-nightly.rb: fixed missing quote

vitorgalvao:
@fanquake Looks like your latest update removed a quote. Perhaps an error on the update script (if you’re already using that)?


vitorgalvao:
@fanquake Looks like your latest update removed a quote. Perhaps an error on the update script (if you’re already using that)?

PR #1391, opened 2015-11-03
Update PyCharm to 5.0

kenzierocks:
If you prefer Community as a separate commit, I can do that too.


vitorgalvao:
> If you prefer Community as a separate commit, I can do that too.

It’d be nice, yes, but is not strictly mandatory. You can do it for future submissions.

PR #1398, opened 2015-11-04
Delete clion-bundled-jdk

andystanton:
JetBrains only offer the version of CLion that comes bundled with their JDK which is now in homebrew-cask


andystanton:
pinging @sebroeder.

PR #1414, opened 2015-11-12
Updated PHPStorm-EAP to 143.790

gcds:


gcds:
Wow that was fast... I just started upgrading by hand :D

adityadalal924:
At homebrew-cask, we aim to please :+1: 

Thanks for the PR!

PR #1428, opened 2015-11-19
Add Scratch 1.4

tjvr:
Hi! I hope this is the right place :-)

The main repo only seems to have Scratch 2.0, which is very different to 1.4. I can't find a 1.4 cask anywhere, so I made this one.


vitorgalvao:
Yes, this is the right place, but why is this version useful?

tjvr:
Scratch 2.0 is a complete rewrite. The older version is easier, and more fun to mod. I'm not sure what to say... but they're definitely pretty different!

PR #1439, opened 2015-11-27
resolve pkg source file error

chuah48263:
```
Downloading http://download.oracle.com/otn-pub/java/jdk/7u80-b15/jdk-7u80-ma
Already downloaded: /Library/Caches/Homebrew/java7-1.7.0_80.dmg
==> Running installer for java7; your password may be necessary.
==> Package installers may write to any location; options such as --appdir are i
Error: pkg source file not found: '/opt/homebrew-cask/Caskroom/java7/1.7.0_80/JDK 7 Update 79.pkg'
```
Solve the above error so that it looks into the newest 'JDK 7 Update 80.pkg'.


fanquake:
Thanks :+1:

PR #1444, opened 2015-12-02
fix invalid header

nabedge:
https://github.com/caskroom/homebrew-versions/commit/e1d1a71e5c7596970542f807e579ea0eb74eef0b

The first line of the commit above was invalid.


fanquake:
In future, please mention the cask name in the commit message.

PR #1446, opened 2015-12-02
Add Go 1.4

mkuzmin:
Go (golang) version 1.4.x

Copied from https://github.com/caskroom/homebrew-cask/blob/master/Casks/go.rb

### DIFF
--- a/Casks/go14.rb
+++ b/Casks/go14.rb
@@ -12,7 +12,7 @@ cask :v1 => 'go14' do
     pkg "go#{version}.darwin-amd64.pkg"
   end
 
-  name 'Go 1.4'
+  name 'Go'
   homepage 'https://golang.org/'
   license :bsd
 
--- a/Casks/go14.rb
+++ b/Casks/go14.rb
@@ -1,0 +1,22 @@
+cask :v1 => 'go14' do
+
+  if MacOS.release <= :snow_leopard
+    version '1.4.2'
+    sha256 '8b8fe5c723663c2da0381643b2f04b0a597f623149095db61818c3fe22be9582'
+    url "https://storage.googleapis.com/golang/go#{version}.darwin-amd64-osx10.6.pkg"
+    pkg "go#{version}.darwin-amd64-osx10.6.pkg"
+  else
+    version '1.4.3'
+    sha256 'cb9eaf5d7be6f0a785937fa6888a8b4e3376abd4c425b5716112da9093ed4094'
+    url "https://storage.googleapis.com/golang/go#{version}.darwin-amd64.pkg"
+    pkg "go#{version}.darwin-amd64.pkg"
+  end
+
+  name 'Go 1.4'
+  homepage 'https://golang.org/'
+  license :bsd
+
+  depends_on :macos => '>= :snow_leopard'
+
+  uninstall :pkgutil => 'com.googlecode.go'
+end


Casks/go14.rb
```diff
@@ -0,0 +1,22 @@
...
+    url "https://storage.googleapis.com/golang/go#{version}.darwin-amd64-osx10.6.pkg"
+    pkg "go#{version}.darwin-amd64-osx10.6.pkg"
+  else
+    version '1.4.3'
+    sha256 'cb9eaf5d7be6f0a785937fa6888a8b4e3376abd4c425b5716112da9093ed4094'
+    url "https://storage.googleapis.com/golang/go#{version}.darwin-amd64.pkg"
+    pkg "go#{version}.darwin-amd64.pkg"
+  end
+
+  name 'Go 1.4'
```
vitorgalvao: The version shouldn’t be part of the name.

vitorgalvao:
Why is this useful?

Casks/go14.rb
mkuzmin: Thanks! Fixed.

mkuzmin:
Not all the products are upgraded to Go 1.5 yet, and version 1.4 is still required in some projects.

vitorgalvao:
In all honesty, I don’t see that as a strong reason, considering that it’s filling a purely hypothetical need and will end up just being an abandoned cask in an already messy repo. If there really is a need for it, I feel it should be an external [tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/brew-tap.md).

I’ve been thinking of some rules for what to include in this repo, and this is actually a good example. Will leave it open for now, as we discuss it.

mkuzmin:
Wow, I didn't expect so strong reaction.
This case is not a purely hypothetical need. I actually need Go 1.4 to build [InfluxDB](https://github.com/influxdb/influxdb/commit/3cab8fac0202b5640674b31997d3bcec30236766) code.
And I was thinking my contribution would help other people.

That's not correct co compare Go with Firefox. It's not a program, it's SDK. You develop in Ruby, so you know reasons why `rvm`, `rbenv`, and `bundler` exist. This is the same story, just in another language.
How does it differ from `java6`, `java7`, python25`, and `python26` packages?

vitorgalvao:
> Wow, I didn't expect so strong reaction.

It wasn’t a strong reaction at all. Either my discourse made it seem stronger than it actually was, in which case I apologise, or you interpreted it stronger than what I’ve written it.

> This case is not a purely hypothetical need. I actually need Go 1.4 to build InfluxDB code.

Exactly, *you*. We cater to a community, not a specific user.

> And I was thinking my contribution would help other people.

But you can’t say for sure, which is why it is hypothetical.

> You develop in Ruby, so you know reasons why `rvm`, `rbenv`, and `bundler` exist.

*Precisely*. A *version manager* is the correct way of handling multiple versions of a language, **not** a *package manager* (which is what homebrew-cask is).

> How does it differ from `java6`, `java7`, `python25`, and `python26` packages?

I’ve repeatedly stated the repo is a mess. Those are part of the problem. Java casks, however, have a very clear purpose: some casks depend on those old java versions.

This is not a personal matter, and this is not a response to your contribution. You contribution is fine and appreciated, but this repo has always been a disorganised mess, and in need of clear rules. It was not your contribution that made me write those now, [another event yesterday was](https://github.com/caskroom/homebrew-cask/pull/15496#issuecomment-161107463).

People are always really vocal about cask refusals, and I honestly do not understand why. In addition, the less a user actually contributed to the project, the more vocal they become. We’re pretty liberal about what to include, and when we refuse something, we have clear reasons. Here, the reason is accepting everything is making this repo a mess, and we need to fix that. It’s all fine and dandy when you submit a cask, but after you do we’re the ones that have to deal with them and manage them, and having too many dead casks makes the job harder.

To be clear, when I say “you”, here, I’m not talking about you specifically, but any submission in general.

**Use taps**. They are, in my view, the absolute best part of homebrew. They allow anyone to leverage the power of homebrew for their specific needs with very little work. The usefulness of your Go contribution will only rapidly decrease over time and the cask will be left here abandoned, adding to the cruft problem. If you believe managing Go in this manner (as opposed to a specialised version manager) is so valuable, why not make a tap for it and manage it yourself?

Again, I want to make clear this is not personal and I’m not judging your contribution (which, once more, is well made and appreciated), but the state of submissions in general, and what we (as a project) can do to fix that.

bdhess:
I don't have a lot of skin in the game, but as an average user I normally expect multiple versions of an SDK to be available in my package manager.  It actually irks me in the infrequent cases where I'm writing Ruby, and need to get rvm, verify and sideload its PGP keys, consult its manpage to refresh my memory on its syntax, and then sideload a Ruby version from a source other than my normal trusted package manager.  I wish I could just `brew install` or `yum install` or whatever.

I think that Ruby, though, is sort of the exception rather than the rule.  For example, homebrew-versions contains multiple versions of GCC, llvm, Perl, and so on as do the central repos for most linux distros.  Heck, there are so many versions of PHP and its associated packages that there's a whole (Homebrew-maintained) tap for them.

fanquake:
@vitorgalvao What are your thoughts for this pull?

vitorgalvao:
Seems like we’re in agreement regarding the new rules, so we should close it. In addition, we’re starting to try to not have duplicates with homebrew, and this [is definitely one](https://github.com/Homebrew/homebrew-versions/blob/master/go14.rb).

PR #1451, opened 2015-12-02
remove daisydisk as snow leopard version is already in main

adityadalal924:
https://github.com/caskroom/homebrew-cask/blob/master/Casks/daisydisk.rb already takes care of this


adityadalal924:
Merged as 36594a2959aaaf6616cd752cfb061b84cb50a428.

PR #1455, opened 2015-12-03
Remove Acorn3

fanquake:


vitorgalvao:
Why remove this one?

adityadalal924:
"Acorn 5 requires 10.10+.  Acorn 4 requires 10.8+.  Both Acorn 2 and 3 require 10.6+.  Acorn 1.5.5 runs on 10.4+.  If you wish to purchase Acorn 1, 3, or 4 go ahead and buy Acorn 5.  Acorn 4.5.6 will accept Acorn 5 registration numbers.  But if you need an Acorn 3 or 1 registration, email support@flyingmeat.com and let us know your registration name and number and we will make you a registration that works with it"

http://flyingmeat.com/acorn/docs/faq.html

vitorgalvao:
That doesn’t explain why (if) we should remove this, though. If you have a license to version 3 and do not wish to buy version 5, this is still useful.

adityadalal924:
Correct, wasn't stating an opinion one way or another, just providing info. Seems like it should stay, as technically 4.x is directly compatible (ie we could add that to the main acorn cask), and acorn3 should stay here in versions, as it's only one version behind.

fanquake:
If the consensus is that this should stay I'll close this.

PR #1458, opened 2015-12-03
Remove Clipmenu Alpha Cask

fanquake:


RJHsiao:
I just realized this cask has been deleted today, but I don't know why it need to be deleted.
Any reason?

adityadalal924:
Dev version was very old:

Download ClipMenu 0.4.4a12
Last updated: May 13, 2010 (JST)
from: http://www.clipmenu.com/dev/

---------
You can still get normal clipmenu -> https://github.com/caskroom/homebrew-cask/blob/master/Casks/clipmenu.rb
(`brew cask install clipmenu`)

RJHsiao:
@adityadalal924 
But Clipmenu Alpha is not equal to Clipmenu Dev. In fact, Clipmenu Alpha is the newest release channel, and ClipMenu 1.0.0a1 has big change than ClipMenu 0.4.4a12.  

ClipMenu 1.0.0a1 release note: http://www.clipmenu.com/blog/2014/11/11/clipmenu-1-0-0a1-released/  
You can see this article is write on Nov. 11, 2014 (JST).

RJHsiao:
BTW, Clipmenu Dev is still exist -- that the real very-old-version you said.
https://github.com/caskroom/homebrew-versions/blob/3b37455c805ee1e405a9644a87e64a4dfa17cb13/Casks/clipmenu-dev.rb

PR #1469, opened 2015-12-03
Remove omnioutliner-professional3 cask

fanquake:


adityadalal924:
Old version is at https://www.omnigroup.com/download/
3 to 4 isn't a free upgrade in some cases, so this should stay

vitorgalvao:
Agreed. This one has a legitimate reason to exist.

PR #1470, opened 2015-12-03
Remove omnigraffle-professional5 cask

fanquake:


adityadalal924:
Old version is at https://www.omnigroup.com/download/
5 to 6 isn't a free upgrade, so this should stay

PR #1472, opened 2015-12-03
Remove VOX Legacy cask

fanquake:


adityadalal924:
I see old versions at http://coppertino.com/old-versions, but the cask would need to be renamed to stay within our naming. (legacy isn't a version name)

vitorgalvao:
I was under the impression vox was gratis, though, so it would fit into the rules of removal.

However, from [the main cask](https://github.com/caskroom/homebrew-cask/blob/master/Casks/vox.rb), it seems like it is freemium after all. I don’t see how on first glance though. Does anyone know if it is indeed freemium?

adityadalal924:
Yeah, I've used it. They added in premium features with the newer versions (radio and some sort of cloud storage service), and also changed the interface, so many people like the old simple, free one.

This cask is probably okay to stay, but should be renamed, as -legacy is quite ambiguous.

vitorgalvao:
Yes, it should stay.

> so many people like the old simple, free one.

So this one in particular should be changed to `:gratis`, then, right?

> but should be renamed

The rule is to use the major version, so should be `vox0`. Looks a bit weird, but it’s accurate.

adityadalal924:
Changes made in https://github.com/caskroom/homebrew-versions/pull/1477

PR #1480, opened 2015-12-04
Add VirtualBox 5.0.8-103449 as virtualbox508103449

jesusfperez:
VIrtualBox 5.0.8-103449

### DIFF
--- a/Casks/virtualbox508103449.rb
+++ b/Casks/virtualbox508103449.rb
@@ -1,0 +1,17 @@
+cask :v1 => 'virtualbox508103449' do
+  version '5.0.8-103449'   # virtualbox.rb and virtualbox-extension-pack.rb should be upgraded together
+  sha256 '4c0987af48340866c3d5902d70f4ef59bdcf05898a36aa425f26c80b29561d4e'
+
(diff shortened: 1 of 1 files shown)


vitorgalvao:
We’re in the process of introducing new rules for this repo.

Virtualbox is one of the problems, actually. Why is this addition useful?

jesusfperez:
Virtualbox has many versions in homebrew-versions but my company need to install virtualbox 5.0.8. So i have created the new formula for it and i thought to upload to official repository. We use cask to install base apps.

vitorgalvao:
This seems like the perfect case for [tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/brew-tap.md)s. Also, if all you need is this, keep in mind we also support loading from a URL, so if you do `brew cask install http://some_location_for_a_cask`, it’ll work.

This way, you don’t need official support from us, we don’t have the repo a mess of abandoned casks (since that is only needed for your company and will likely be abandoned here when you no longer have a use for it), and you don’t have to wait for us to approve and merge casks, just update your own definitions/repos.

Everyone wins.

Issue #1450, opened 2015-12-02
Rules for what to include in this repo


vitorgalvao:
Pinging @caskroom/maintainers.

To put it bluntly, this repo is a dump. While thinking of how we could organise it better, I though we perhaps needed a few more repos, but this actually has a small number of casks (under 350), so we’re well in time to make it manageable.

What I see as part of the issue is how just about any old version gets included, for any reason. The recent [submission of an old version of Go](https://github.com/caskroom/homebrew-versions/pull/1446) is a good example. This has [marginal hypothetical benefits and will end up being a neglected cask](https://github.com/caskroom/homebrew-versions/pull/1446#issuecomment-161432754), and I believe we should not include it. This is also true for a truly ridiculous amount of [old firefox versions](https://github.com/caskroom/homebrew-versions/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+Add+Firefox+version). For most of these, [there isn’t anything close to a significant demand for them or clear advantage to including them and hence should not be officially supported by us](https://github.com/caskroom/homebrew-versions/pull/1346#issuecomment-148812007).

My proposal for the rules of cask inclusions in this repo is as follows (following [our nomenclature](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#finding-a-home-for-your-cask)):

+ Include the latest minor version of legacy versions of commercial and freemium software<sup>1</sup>.
+ Legacy versions of commercial and freemium software are restricted to a maximum of five casks.
+ Include beta, development, unstable, nightly, early access program, ….
+ Include language and regional versions other than US english.
+ Include alternate versions (Firefox Developer Edition, Community Editions, Pro versions).
+ Refuse legacy versions of gratis or open-source software, unless there is a clear demonstrable need for them.
+ Legacy versions of gratis or open-source software that were accepted should be removed after one year.
+ Include casks that do not fit the rules, but need to exist somewhere since they are required by other casks.

Some reasonings:

[...]

PR #1503, opened 2015-12-18
Add older quickbooks desktop editions

sscotth:


vitorgalvao:
Thank you, but now. this is precisely the type of abandoned submission casks we’re trying to prevent with https://github.com/caskroom/homebrew-versions/issues/1450.

sscotth:
Can you explain further?

In the realm of commercially licensed software it is similar to the adobe suite about needing older versions, but even more so because the files are not forwards compatible.

My accountant client has to maintain several older versions of the software because she cannot work on her client's files created with the old version and maintain its compatibility. i.e. A company file created with v2012 can be opened by v2013, but making changes requires you to "upgrade" the file to v2013 making it inoperable with v2012 in the future.

I looked at #1450 and cask seems to fit
* Legacy versions of commercial and freemium software are restricted to a maximum of five casks. (I dropped v2009 and v2010 to conform)
* There is a clear demand and reason to include legacy versions of commercial software — you don’t want to or can’t spend the money to get the latest version.

vitorgalvao:
Also from that issue

> there isn’t anything close to a significant demand for them or clear advantage to including them and hence should not be officially supported by us.

Also, from [a comment further down](https://github.com/caskroom/homebrew-versions/issues/1450#issuecomment-163554990), if we follow the link, we’ll find:

> Versions is not intended to be used for all and any old versions you personally require for xyz project; formulae submitted here should be expected to be used by a reasonable number of people and supported by contributors long-term.

There isn’t a clear demand for this, and it seems like you’re the only one that needs it. Like you said on the main repo, they have been selling this for fifteen years, and it’s in so low demand for our users, just today was the current version even submitted.

Granted, just by skimming that issue, you couldn’t have known all that, but I am this very moment polishing those rules for inclusion, so they should be up in a while.

In this case, (and taking again from the link in that comment) the solution is:

> You may wish to consider hosting your own [tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) for formulae you wish to personally support that do not meet the above standards.

Also keep in mind these rules are very knew and may change. At the moment, though, the goal is to make this repo less of a dump, so for the near future we will be harder on refusals, while we sort everything out.

sscotth:
I understand and agree with your updated rules. I also agree that the demand for this cask has not presented itself until now. However, I do think having one or two prior versions here would be appropriate, especially when not everyone upgrades for $200+ the moment a new edition comes out.

If you do not agree, I will maintain a tap.

vitorgalvao:
> I also agree that the demand for this cask has not presented itself until now.

Yes, but that is a demand of a single person, so it does not really fit the rule.

[Having said that](http://youtu.be/ENHVQ2gslp8), it is indeed a simple cask, so lets compromise on two old versions (`2015` and `2014`). Would that be enough for you?

If this really will be used by a reasonable number of people *or*<sup>1</sup> be supported by contributors long-term, then we can expect that when new versions come out (presumably `2017` and `2018`), both the main cask and the ones here will be updated. They’ll also have the dates in the name, so should be easy to spot. If they’re not, we can handle them then.

If we’re in agreement, please open a new issue and @mention me. Also, the casks will need a few fixes (already made to the main one). I’ll pick one of the ones currently here and comment on it.

---

<sup>1</sup> The rule is *and*, but since we’re still trying it out, lets see in this case.

Casks/quickbooks-desktop-2015
```diff
@@ -0,0 +1,12 @@
+cask 'quickbooks-desktop' do
+  version '2015'
+  sha256 :no_check
+
+  url "http://http-download.intuit.com/http.intuit/Downloads/#{version}/Latest/QuickBooksProMac#{version}.dmg"
+
```
vitorgalvao: Please remove this empty line.

Casks/quickbooks-desktop-2014
```diff
@@ -0,0 +1,12 @@
+cask 'quickbooks-desktop' do
+  version '2014'
+  sha256 :no_check
+
+  url "http://http-download.intuit.com/http.intuit/Downloads/#{version}/Latest/QuickBooksProMac#{version}.dmg"
+
```
vitorgalvao: Please remove this empty line.

Casks/quickbooks-desktop-2014
```diff
@@ -0,0 +1,12 @@
+cask 'quickbooks-desktop' do
```
vitorgalvao: This needs to be `cask 'quickbooks-desktop-2014' do`.

Casks/quickbooks-desktop-2014
```diff
@@ -0,0 +1,12 @@
...
+  sha256 :no_check
+
+  url "http://http-download.intuit.com/http.intuit/Downloads/#{version}/Latest/QuickBooksProMac#{version}.dmg"
+
+  name 'QuickBooks Desktop'
+  homepage 'http://quickbooks.intuit.com/mac/'
+  license :commercial
+
+  app "QuickBooks #{version}.app"
+end
```
vitorgalvao: It has a “no newline at end of file” warning. Those are [actually meaningful](http://unix.stackexchange.com/a/18789).

Casks/quickbooks-desktop-2014
```diff
@@ -0,0 +1,12 @@
+cask 'quickbooks-desktop' do
+  version '2014'
+  sha256 :no_check
```
vitorgalvao: These versions presumably are no longer updated, right? So they should have a `sha256`.

sscotth: I'm not sure how often the are updated, I will add the sha and if it becomes an issue we can remove it.

sscotth: Could I comment the version like this: `version '2014' # 15.0.9 R10`

Casks/quickbooks-desktop-2014
jawshooah: Shouldn't it be `quickbooks-desktop2014`?

sscotth: I'm not sure what the standard is.

microsoft-office-2011
adobe-photoshop-cc
adobe-photoshop-cs5

jawshooah: @vitorgalvao I wouldn't call `2014` part of the actual name of the software, and we're certainly treating it like a version number. What do you think?

vitorgalvao: > Shouldn't it be `quickbooks-desktop2014`?

Yes, you’re right.

> I wouldn't call `2014` part of the actual name of the software, and we're certainly treating it like a version number.

Yes, I was thinking of it solely as the version number, since that’s both what we have to go by *and* is what anyone will expect.

Casks/quickbooks-desktop-2014
vitorgalvao: Lets add shasums for now, and remove them later if it becomes a problem.

Version should still be `2014` without a comment, since people won’t likely know what to look for otherwise.

PR #1504, opened 2015-12-18
readme: add acceptable casks

vitorgalvao:


README.md
```diff
@@ -17,5 +17,24 @@ You can now install alternate versions of Casks.
...
+
+**Please note that `caskroom/versions` is currently in the process of major changes in what we support, how long for and on what basis. For that reason, we may be harder on refusals for the near future.**
+
+[Similarly to `homebrew-versions`](https://github.com/Homebrew/homebrew-versions#acceptable-formulae), `caskroom/versions` is not intended to be used for all and any old versions you personally require for xyz project; casks submitted here should be expected to be used by a reasonable number of people and supported by contributors long-term.
+
+You may wish to consider hosting your own [tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) for casks you wish to personally support that do not meet the above standards.
+
+For this repo, rules for consideration are (following our [nomenclature](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#finding-a-home-for-your-cask)):
+
++ Include the latest minor version of legacy versions of commercial and freemium software<sup>1</sup>.
```
jawshooah: Where's the footnote?

vitorgalvao: Forgot about that. Initially was going to include it, but then decided against it. Removed the marker.

bdhess:
Just in general, and perhaps the question is tangential to this PR, how do you feel about something like [Popularity Contest](http://popcon.debian.org) and developing minimum usage targets to decide when to retire a cask rather than hard limits on number of previous versions or duration for which a Cask should be maintained?  That said, I appreciate that you are trying to clarify rules and move things forward immediately, and I'm not trying to stand in the way of that.

vitorgalvao:
In general, [I’m against collecting user data](https://github.com/caskroom/homebrew-cask/issues/4323)<sup>1</sup>. Reasons being collected data can be done in two ways, opt-in or opt-out. If it is done by opt-out, it is a privacy violation (however small), but if it is done opt-in, the data is pretty much useless. The solution is, then, to not waste time implementing that.

[Having said that](http://youtu.be/ENHVQ2gslp8), those arguments are in relation to a popularity contest in the sense of “which apps are more popular”. In the case you mention, it’s just about “is anyone even downloading this app”, which I could see a better case for. Even if a few number of users opts-in and we ignore apps that don’t have “hits”<sup>2</sup>, we could still consider the ones that have them and would be removed otherwise, as apps we should keep.

I’m still not too comfortable with the idea, though. How would we do it? Once you have homebrew-cask installed, you don’t need to have any more contact with us. Everything happens on your machine, and you call the app servers directly when you need to install something.

This would require we either keep a list of everything you installed and periodically send it to us (yuck! Also, goes against [our new mode of thinking](https://github.com/caskroom/homebrew-cask/issues/13201) where we try to emulate a manual installation as closely as possible) or that we, for instance, send a hit to a server somewhere whenever you run a `brew cask install`.

All of those require we maintain and pay for a server somewhere for this (not to mention build and maintain the code for this in homebrew-cask itself) and I simply cannot see how it’d be worth it, particularly right now when we have real functional issues to solve.

Lastly, since we’ll be joining up with homebrew, this also seems like something weird to have when they wouldn’t, particularly since we’ll try to reuse their code as much as we can. For this to be implemented, seems like it would need to be a discussion and implemented on the core.

---

<sup>1</sup> Not recommending you read through the issue, though, just giving it as context in case you do want to read my points.

<sup>2</sup> As in “were we hit with a request for it”.

adityadalal924:
LGTM.

PR #1530, opened 2015-12-29
Delete totalfinder-beta.rb

Zirro:
v1.7.10 is already stable and development appears to have ceased since
the introduction of SIP in El Capitan.


vitorgalvao:
Thank you.

PR #1532, opened 2015-12-30
Add Unity v5.2.3f1

cybertk:


vitorgalvao:
Is this a beta?

cybertk:
It's a old version of Unity, but widely used.

vitorgalvao:
This [fits our rules](https://github.com/caskroom/homebrew-versions#acceptable-casks), but lets refuse it for now. We’re still trying to clean the repo, and if we merge this now we may end up removing it anyway. Betas as fine, but for old versions lets go slow for now.

PR #1542, opened 2016-01-05
Delete a bunch of casks

vitorgalvao:
Mostly old versions and things that don’t even exist anymore. I’ve left a few that I am certain people are still using.

Cleaning this up like this in terms of old versions means we can now start to apply the rules.


adityadalal924:
Office 2011 and maybe parallels10 should stay, AFAIK

vitorgalvao:
> Office 2011 and maybe parallels10 should stay, AFAIK

Included.

adityadalal924:
Also, thoughts on the omni- casks? Upgrades aren't free but I don't use them so I have no frame of reference to know if people use the older versions. 

I guess we can always delete but selectively restore if there is demand.

vitorgalvao:
> I guess we can always delete but selectively restore if there is demand.

That’s my idea. There are more old commercial apps here that are being deleted under the same though.

metyl:
All old version of Idea are still [available] (https://confluence.jetbrains.com/display/IntelliJIDEA/Previous+IntelliJ+IDEA+Releases), can we have it back? People are still using them.

philadams:
Yikes, this was a frustrating commit. The whole purpose of this repo is to make installing older versions of software super straightforward. Came to install some old software on a new machine... and the software in a PR that I'd had merged into this repo > 3 years ago is gone!

Guess we're back to manually installing things?

jawshooah:
The reasons for removing casks from the repo are made clear in the readme. If you still need something that used to be here, you're free to create a custom tap for it. I'll never understand how users always forget about that feature...

jawshooah:
Alternatively, if you really feel that the software you need belongs here *and meets the requirements listed in the readme*, you're also free to submit a PR re-adding it with adequate justification.

philadams:
Understood - thanks.

PR #1543, opened 2016-01-05
Bump webstorm eap to v12, fix 404

paulirish:
Current path is 404ing.  This PR fixes that by upgrading to the recently released (dec 29th) eap v12.

Source: https://confluence.jetbrains.com/display/WI/WebStorm+EAP


vitorgalvao:
Thank you.

PR #1578, opened 2016-01-15
Update clipmenu-alpha.rb for bug fixed

RJHsiao:
The error message below will displayed when trying to run `brew cask list` and `brew cask install clipmenu-alpha` before bug fixed:  
````
Error: Cask 'clipmenu-alpha' definition is invalid: Bad header line: '{:v1=>"clipmenu-alpha"}' does not match file name
````


vitorgalvao:
Thank you. This one slipped through.

PR #1579, opened 2016-01-16
Fixed appcast in osxfuse-beta

tsparber:


zchee:
@tsparber Ah, I missed :(
Thanks!

PR #1596, opened 2016-01-24
Create eclipse-java-luna.rb

stvhwrd:
Some Eclipse plugins (eg. [Google Apps Script](https://developers.google.com/eclipse/?hl=en)) are only compatible with older versions of Eclipse.  Handy to have the option to install older versions.


vitorgalvao:
As [per our rules](https://github.com/caskroom/homebrew-versions#acceptable-casks), and since this is open source, we’ll refuse it.

You may wish to add it to your own tap.

PR #1668, opened 2016-02-06
updated sublime-text-dev (3101)

ravecode:


Casks/sublime-text-dev.rb
```diff
@@ -1,6 +1,6 @@
 cask 'sublime-text-dev' do
-  version '3100'
-  sha256 '2612f254f7ba055129309e2865d979aeb09e84ee6ace889e4a94e79b978f2af9'
+  version '3101'
+  sha256 '590004d52b05525e32e1e0371b1aa49aa1c8e8173fa5f1d9cd28e4c60a5e8c53'
 
   url "https://download.sublimetext.com/Sublime%20Text%20Build%20#{version}.dmg"
   appcast 'https://www.sublimetext.com/updates/3/dev/appcast_osx.xml',
```
vitorgalvao: Updating the `:checkpoint` value in `appcast` is required. You can find it with `curl --silent --compressed "{{appcast_url}}" | sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256 | awk '{ print $1 }'`.

Alternatively, you may wish to use [`cask-repair`](https://github.com/vitorgalvao/tiny-scripts/blob/master/cask-repair). It can at once update the cask’s version (according to what you give it) taking care of updating both the cask’s `sha256` and the appcast’s `:checkpoint`, while still giving you the chance to perform other manual fixes.

You can easily install it with `brew install vitorgalvao/tiny-scripts/cask-repair`. `cask-repair --help` will show you some options, but as a whole the script will guide you through the update process.

ravecode:
This PR was generated using `cask-repair ` so the appcast checkpoint should have been updated automagically, however, it appears that the info at the appcast url is no longer in sync with the current version of `sublime-text-dev`.

vitorgalvao:
Seems like it. Lets see if it continues like so. Would you mind contacting them about it (perhaps something broke in their build automation)?

ravecode:
I got in contact and the appcast.xml was supposedly removed from the build process due to it not being used in ST3 for a long time, however they will be re-adding it back in the next build.

vitorgalvao:
Good news. Thank you or contacting them.

PR #1670, opened 2016-02-07
Update MacVim-Kaoriya to 20160205

delphinus35:
ref https://github.com/splhack/macvim-kaoriya/releases/tag/20160205


delphinus35:
sorry, I found some bugs on this formula or this version of macvim. once I close this.

delphinus35:
I found it cannot have `,` character in `version`. In vim, the `runtimepath` setting have many paths separated by `,`. If each path have `,` in itself, vim cannot determine paths in `runtimepath` correctly. So this PR use `:` instead of `,` to avoid this.

vitorgalvao:
After https://github.com/caskroom/homebrew-cask/issues/13201, we can revert this.

PR #1671, opened 2016-02-07
Add Zulu 7

nathansgreen:
Azul's supported line of OpenJDK 7 builds.

### DIFF
--- a/Casks/zulu7.rb
+++ b/Casks/zulu7.rb
@@ -1,32 +1,42 @@
 cask 'zulu7' do
-  version '1.7.0_95,7.13.0.1'
+  version '7.0_95,7.13.0.1'
   sha256 '3744b75e621876934c3dc256491d5a3e58b73cd96e2b809c77a333f7e34debbe'
 
-  url "http://cdn.azul.com/zulu/bin/zulu#{version.after_comma}-jdk#{version[2..7].underscores_to_dots}-macosx_x64.dmg",
+  url "http://cdn.azul.com/zulu/bin/zulu#{version.after_comma}-jdk#{version.before_comma.underscores_to_dots}-macosx_x64.dmg",
       referer: 'http://www.azul.com/downloads/zulu/zulu-mac/'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
In [acceptable tasks](https://github.com/caskroom/homebrew-versions#acceptable-casks):

adidalal:
@vitorgalvao Should be the same as having old versions of Java, honestly (which we do allow). Some programs need them.

vitorgalvao:
@adidalal Alright, will leave the `zulu` ones to you.

Casks/zulu7.rb
```diff
@@ -0,0 +1,32 @@
+cask 'zulu7' do
+  version '1.7.0_95,7.13.0.1'
+  sha256 '3744b75e621876934c3dc256491d5a3e58b73cd96e2b809c77a333f7e34debbe'
+
+  url "http://cdn.azul.com/zulu/bin/zulu#{version.after_comma}-jdk#{version[2..7].underscores_to_dots}-macosx_x64.dmg",
```
vitorgalvao: @nathansgreen Please change the interpolation here, though. See [the other `zulu`](https://github.com/caskroom/homebrew-cask/blob/cf459cdb18ed3d772d06b3976ddb045213981fd9/Casks/zulu.rb#L5).

nathansgreen:
The Zulu build is actually newer than Oracle's builds and seems to be updated more frequently.

nathansgreen:
@vitorgalvao, instead of adding a `caveats` section like Zulu, should I just use `zap`? I'm not sure if that's an appropriate use of zap.

vitorgalvao:
`zap` is for configurations. `caveats` should be used.

adidalal:
Thank you for the contribution. It needed some fixes, so they were made in 2293df997f18ae52053abcb6592ded6cb8a1c093. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

PR #1718, opened 2016-02-11
updated bitbar-beta (1.5.0-beta2)

tresni:


tresni:
Side note, I ran into Caskroom/homebrew-cask#16475 trying to do this.  For me the solution was simply to execute:

```sh
mkdir /Library/Caches/Homebrew/Casks
```

Then everything worked as expected.  (The folder did not exist prior and nothing was attempting to create it it seems.)

PR #1729, opened 2016-02-14
Update the destination of the symlink

ffittschen:
The destination of the symlink should be in CamelCase, like the source. In case the stable and the beta versions are installed next to each other, I added a "Beta" to the name.

### DIFF
--- a/Casks/bitbar-beta.rb
+++ b/Casks/bitbar-beta.rb
@@ -9,5 +9,5 @@ cask 'bitbar-beta' do
   homepage 'https://github.com/matryer/bitbar/'
   license :mit
 
-  app 'bitbar.app'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
You misunderstanding how it works. `app` is the source, by changing it to a value that isn’t real, you’re breaking the cask. What you’re looking for is [`target:`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#renaming-the-target), but [we don’t accept this types of renaming](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases) on behalf of the user.

PR #1773, opened 2016-03-02
airmail-beta has no versioned download

nstCactus:
Revert of commit [2723399](https://github.com/caskroom/homebrew-versions/commit/2723399480fe4957cef1e3bd805adb4a882aca84) which breaks the formula on every Airmail Beta update (sha mismatch)


vitorgalvao:
This was explained on the commit.

PR #1793, opened 2016-03-05
Update Airmail Beta to version 2.6.1,353

mauricerkelly:
This commit updates the version, sha256 and checkpoint stanzas. It
also changes the license from `:unknown` to `:commercial`

### DIFF
--- a/Casks/airmail-beta.rb
+++ b/Casks/airmail-beta.rb
@@ -1,14 +1,14 @@
 cask 'airmail-beta' do
-  version '2.6.1,352'
-  sha256 '8ddc915a1e7e5031ff592925d55d8016992e17f497d90135d4dbafac127ee6f3'
+  version '2.6.1,353'
+  sha256 'a728b354f1065f2b3d92f053af42e5288546103522438f39e5f8ad210442fb29'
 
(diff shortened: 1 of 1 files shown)


mauricerkelly:
@vitorgalvao I'm guessing a little here. I updated the version and sha256 stanzas, and I computed an updated checkpoint based [on the instructions here](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md)

I'm ignorant though as to what `brew cask` actually uses the checkpoint for? When a new Airmail Beta version was released the existing cask became immediately broken (the sha256 did not match). I would have expected the explicit versioning (`:version`/`:checkpoint`) to pin the cask to downloading a specific version, but it didn't.

Unlike a lot of beta/nightly casks I've been updating in the past, this cask uses the same download URL for everything, and as a result becomes invalid as soon as a new version is released. This makes it really inconvenient to use given the frequency Airmail Beta is updated - Airmail Beta is particularly troublesome as it provides no in-app update - every new beta requires an update to the cask.

vitorgalvao:
> I'm ignorant though as to what `brew cask` actually uses the checkpoint for?

It’s part of [a set of scripts](https://github.com/caskroom/homebrew-cask/pull/19274) to make catching outdated casks easier.

As we make those better, we hope to catch these cases faster and faster. For now, these problems are actually good to catch the flaws in the system.

PR #1801, opened 2016-03-09
Add Firefox 38 ESR

tjanson:
Mozilla intends to support Firefox Extended Support Releases (ESR) for 9 major versions [(proposal)](https://wiki.mozilla.org/Enterprise/Firefox/ExtendedSupport:Proposal), while releasing an ESR every 7 majors. I.e., there is a two version overlap where two ESRs are supported.

Firefox 38 ESR will thus still be supported for two more versions (45 ESR was released yesterday). I think it's in the interest of Homebrew users and in the spirit of Mozilla's ESR plan to keep Casks of supported ESRs around.

FF 38 ESR & this cask are expected to retire 2016-06-07.

### DIFF
--- a/Casks/firefox-38esr.rb
+++ b/Casks/firefox-38esr.rb
@@ -1,0 +1,16 @@
+cask 'firefox-38esr' do
+  version '38.7.0'
+  sha256 '9bb953b21f54559b1d25bf4f6026108d54151376f6895dd9c39bb738e295b90b'
+
+  url "https://download-installer.cdn.mozilla.net/pub/firefox/releases/#{version}esr/mac/en-US/Firefox%20#{version}esr.dmg"
+  name 'Mozilla Firefox'
+  homepage 'https://www.mozilla.org/en-US/firefox/organizations/'
+  license :mpl
+
+  app 'Firefox.app'
+
+  zap delete: [
+                '~/Library/Application Support/Firefox',
+                '~/Library/Caches/Firefox',
+              ]
+end


vitorgalvao:
[Acceptable casks](https://github.com/caskroom/homebrew-versions#acceptable-casks).

tjanson:
Just in case I wasn’t clear, here are the two ways homebrew-versions could handle Firefox ESR:

* Include the latest ESR only, effectively shortening the ESR support cycle (for homebrew users) from 9 major versions to 7.
* Include the ESRs that Mozilla supports (e.g., currently 38 and 45), retiring the older version in parallel with Mozilla.

I strongly favor the second option and ask you to reconsider (though I’ll respect your decision). In my eyes, the entire point of ESR to have long term support for a single version, and by choosing the first option, we shorten this support for our users by 3 months. If homebrew-versions includes Firefox ESR at all, it should in my opinion shadow Mozilla’s schedule.

If it helps, I volunteer to keep an eye on the cask(s) and keep it up to date.

tjanson:
On second thought, I think it would be best *not* to have a `firefox-esr` cask at all, but rather have version-specific casks (`firefox-38esr` and `firefox-45esr` and so on), which are introduced and retired in parallel with Mozilla.

The reasoning is as follows: If you’ve chosen to install Firefox 38 ESR, you likely want to stay on that major version until Mozilla ends its support, and in particular you may not want to update to 45 ESR upon release.

Specifically, this would require the following work:

* Mozilla’s support cycle lasts a bit over a year, i.e., approximately once a year a new ESR cask would need to be created and an old cask be removed. (E.g., the next two actions would be removing 38 ESR in June, then introducing 52 ESR in 2017Q1.)
* When overlapping, both ESR casks would need to receive minor version bumps.

I understand someone needs to commit to doing in order to avoid a “graveyard” of truly outdated versions. But I think it is very worthwhile for Firefox ESR, and again I’m willing to do this work.

PS. If you prefer, I’ll move this to a separate issue.

vitorgalvao:
> I understand someone needs to commit to doing this in order to avoid a “graveyard” of truly outdated versions

That’s the big issue, and one of the reasons this repo received a big clean, recently. I don’t believe there’s enough demand for these to justify the effort, to be honest.

> and again I’m willing to do this work

And by doing so, you’ll always be beholden to when/how the homebrew-cask project merges those changes. It’ll also increase the number of casks by a ridiculous amount, one for each language.

Since you’re both the one interested in this and offering to do the legwork, I recommend you make your own [tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md).

tjanson:
Okay. I still think extended release versions like this would be one of the more useful additions to homebrew-versions, but I understand your concern. Thanks for all your work, I appreciate it.

In case anyone needs it, I did create a tap: [tjanson/firefox-esr](https://github.com/tjanson/homebrew-firefox-esr).

PR #1859, opened 2016-03-28
RubyMine-EAP 2016-RC1

yurikoles:


yurikoles:
We need to fix some paths in zap stanza.

PR #1876, opened 2016-04-02
Update CLion EAP (2016.1.1)

tjanson:


Casks/clion-eap.rb
```diff
@@ -1,19 +1,27 @@
...
+                "~/Library/Preferences/CLion#{version.major_minor}",
+                "~/Library/Application Support/CLion#{version.major_minor}",
+                "~/Library/Caches/CLion#{version.major_minor}",
+                "~/Library/Logs/CLion#{version.major_minor}",
               ]
+
+  caveats <<-EOS.undent
+    Please manually change to the EAP update channel via:
+       Preferences > Appearance & Behavior > System Settings > Updates
+    EOS
```
vitorgalvao: Too indented.

vitorgalvao: There should also be a message warning why the caveat exists in the first place, and when it should be removed (because it is very much temporary).

tjanson: Can you clarify what you mean by too indented? Do you mean the second line? That was on purpose as a kind of blockquote, but I can give remove that if you prefer.

As for the explanatory message: So I should add a "This is a stable release and there is currently no EAP release, yada yada"?

vitorgalvao: I mean the `EOS` line itself. it should be on the same level of `caveats <<-EOS.undent`(i.e. it has two spaces too many).

> "This is a stable release and there is currently no EAP release, yada yada"

Yes, whatever you think is fine. Just clarify that the `caveats` should eventually be removed, and *why*.

tjanson: Oh, okay, should that EOL be changed in [this test cask, too](https://github.com/caskroom/homebrew-cask/blob/master/test/support/Casks/with-caveats.rb)? (That’s what I used as a template.)

vitorgalvao: > should that EOL be changed in [this test cask, too](https://github.com/caskroom/homebrew-cask/blob/master/test/support/Casks/with-caveats.rb)? (That’s what I used as a template.)

Yes, please, if you don’t mind. Link back to this issue, so if another maintainer sees it before I do, they have some context.

tjanson:
All updated with more detailed message and comment about removal. Let me know if there is anything else.

PR #1877, opened 2016-04-02
Updated Datagrip EAP (2016.1 EAP, build 145.735.3)

tjanson:
The files

* `~/.DataGrip2016.1`
* `~/Library/Preferences/jetbrains.datagrip.*.plist`

are not present on my machine, so I cannot confirm their spelling / location.
(The other 4 zap entries are present; note the camelcase.)

### DIFF
--- a/Casks/datagrip-eap.rb
+++ b/Casks/datagrip-eap.rb
@@ -1,11 +1,20 @@
 cask 'datagrip-eap' do
-  version '145.404.2'
-  sha256 'c83131497cd8d75acaa001b9dd4a9b602b6c09ce29d912cde9ddb908f4bd245c'
+  version '2016.1,145.735.3'
+  sha256 '1307ec1722361c8777f1f2f7a57977e246b161d5969ce24d3c0caabdb7212370'
 
-  url "https://download.jetbrains.com/datagrip/datagrip-#{version}.dmg"
+  url "https://download.jetbrains.com/datagrip/datagrip-#{version.after_comma}.dmg"
   name 'Datagrip EAP'
   homepage 'https://confluence.jetbrains.com/display/DBE/DataGrip+2016.1+EAP'
   license :commercial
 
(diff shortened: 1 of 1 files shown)


Casks/datagrip-eap.rb
```diff
@@ -1,11 +1,20 @@
...
+  sha256 '1307ec1722361c8777f1f2f7a57977e246b161d5969ce24d3c0caabdb7212370'
 
-  url "https://download.jetbrains.com/datagrip/datagrip-#{version}.dmg"
+  url "https://download.jetbrains.com/datagrip/datagrip-#{version.after_comma}.dmg"
   name 'Datagrip EAP'
   homepage 'https://confluence.jetbrains.com/display/DBE/DataGrip+2016.1+EAP'
   license :commercial
 
-  app 'Datagrip EAP.app'
+  app "Datagrip #{version.major_minor_patch} EAP.app"
```
vitorgalvao: Do you want ` #{version.after_comma.major_minor_patch}`?

tjanson: No, it should be working as intended – the desired string is "Datagrip 2016.1 EAP.app", and when *2016.1.1* is released, it will likely be "Datagrip 2016.1.1 EAP.app" if PyCharm is any indication.

vitorgalvao: Then why not just `#{version.before_comma}`?

tjanson: Good point, though it’s equivalent here and `major_minor_patch` is more explicit (since in contrast to the zap stanza the minor is not omitted here). I’ll change it if you prefer.

vitorgalvao: I see your point, but disagree. Seeing `before` or `after` is more immediate when you’re auditing the cask.

> I’ll change it if you prefer.

Please do, yes.

tjanson:
Updated as per comments.

PR #1889, opened 2016-04-07
Update phpstorm-eap to 145.844.5

deizel:
http://blog.jetbrains.com/phpstorm/2016/04/phpstorm-2016-1-1-update-eap-145-844/


tjanson:
This breaks `#{version.major_minor}` and thus the zap stanza. One way to fix this is to have both the release version (2016.1.1) and build number (145…) in `version`, separated by a comma [like I did here](https://github.com/caskroom/homebrew-versions/commit/0255e8aabcc0c12d53b4dae977705fbb55cadfd6) (better alternatives welcome).

(I will submit a fix in the next few days if still needed.)

PR #1899, opened 2016-04-09
Update Pycharm EAP (Pro & CE) to 2016.1.2 (stable!)

tjanson:
Jetbrains has again released a stable that supersedes the last EAP release, similarly to CLion etc. at version 2016.1.
That's a bit of a shame that this is becoming common, since it requires the user to manually change the update channel, which makes the EAP casks less useful.

----

If this keeps happening, it might be worth checking whether the channel can be changed programmatically.

Update: It looks like it’s as simple as adding

```xml
    <option name="UPDATE_CHANNEL_TYPE" value="eap" />
```

to `~/Library/Preferences/PyCharm#{version.major_minor}/options/updates.xml`.

That seems pretty feasible and non-destructive. (I see [something similar has been done before to localize Adobe software](https://github.com/caskroom/homebrew-versions/blob/f98ada9a40c7b2e3fe52a8c63cc3be3e62c5fbcc/Casks/adobe-illustrator-cc-de.rb#L14-L20).) What do you think?

### DIFF
--- a/Casks/pycharm-ce-eap.rb
+++ b/Casks/pycharm-ce-eap.rb
@@ -1,15 +1,15 @@
 cask 'pycharm-ce-eap' do
-  version '2016.1.1,145.597.11'
-  sha256 '2fa86d695a2f4f161ea84a9c215de2b3de2cf054fa9ed6493fd9b8a228aab8c5'
+  version '2016.1.2'
+  sha256 'e68b9fa401500311bcc3cbbcbd6b3ab02849dcac3bdb6864389aec17c6c7e992'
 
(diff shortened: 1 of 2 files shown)


vitorgalvao:
> That seems pretty feasible and non-destructive. (I see [something similar has been done before to localize Adobe software](https://github.com/caskroom/homebrew-versions/blob/f98ada9a40c7b2e3fe52a8c63cc3be3e62c5fbcc/Casks/adobe-illustrator-cc-de.rb#L14-L20).) What do you think?

Sure. Do some tests with different EAP casks, and if it works, we can consider it.

PR #1903, opened 2016-04-12
Upgrade Firefox (localized version) to 45.0.2

mahori:
[view changes](https://www.mozilla.org/en-US/firefox/45.0.2/releasenotes/)


vitorgalvao:
Thank you.

PR #1956, opened 2016-04-30
Don't track version of airmail-beta

yurikoles:
Revert recent changes tracking version.


vitorgalvao:
This has an `appcast`. All casks with them are versioned.

PR #2045, opened 2016-05-16
updated android-studio-canary (2.1.1.0,143.2821654)

tjanson:
#### Editing an existing cask


tjanson:
I see that this cask is missing the verified download host comment. (The stable version of the cask over in `homebrew-casks` [has the comment](2.1.1.0,143.2821654).)

Also, this cask suffers from the same issue as all the other IDEA IDEs: It’s currently a stable release and by default uses the stable update channel rather than – in this case – `canary`.

I did not adress those issues here. But this update [fixes a security vulnerability](https://plus.google.com/103342515830390186255/posts/5TGKhcwtip5).

vitorgalvao:
Not having a comment is correct here. Doesn’t need it due to the same domain.

PR #2066, opened 2016-05-19
Update appcode-eap to latest '145.1503.11'

eunikolsky:

### DIFF
--- a/Casks/appcode-eap.rb
+++ b/Casks/appcode-eap.rb
@@ -12,8 +12,6 @@ cask 'appcode-eap' do
   app 'AppCode 2016.1.3 EAP.app'
 
   zap delete: [
-                "~/.Appcode#{version.major_minor}",
-                # TODO: expand/glob for '~/Library/Preferences/jetbrains.appcode.*.plist',
                 "~/Library/Preferences/AppCode#{version.major_minor}",
                 "~/Library/Application Support/AppCode#{version.major_minor}",
                 "~/Library/Caches/AppCode#{version.major_minor}",
--- a/Casks/appcode-eap.rb
+++ b/Casks/appcode-eap.rb
@@ -1,6 +1,6 @@
 cask 'appcode-eap' do
-  version '2016.1.2'
-  sha256 '24188b9b99f2d5afa85665711b33cf8cfc439916d5f822e41d7dba31115060c9'
+  version '145.1503.11'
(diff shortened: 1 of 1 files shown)


Casks/appcode-eap.rb
```diff
@@ -9,23 +9,12 @@
 
   conflicts_with cask: 'appcode'
 
-  app 'AppCode.app'
+  app 'AppCode 2016.1.3 EAP.app'
 
   zap delete: [
-                "~/.Appcode#{version.major_minor}",
```
vitorgalvao: Why remove these?

eunikolsky: I've never seen those directories created by `appcode`. both were added once here: 5a44a163a008f92d1f02f0ab189d2f62a9c0de7d and i'm not sure why

vitorgalvao: Pinging @tjanson, then.

tjanson: I have no compelling reason other than that they are present in [the stable cask](https://github.com/caskroom/homebrew-cask/blob/master/Casks/appcode.rb#L15), which I used as reference. A similar entry is in many of the IDEA IDEs (e.g., [Rubymine](https://github.com/caskroom/homebrew-cask/blob/master/Casks/rubymine.rb#L15). I don’t know whether that’s a remnant from some older version or still relevant.

tjanson: Actually, I found an [official source](https://www.jetbrains.com/help/idea/2016.1/directories-used-by-intellij-idea-to-store-settings-caches-plugins-and-logs.html?origin=old_help#d152420e53) which says that this directoy exists on "Linux and the other UNIX systems", but in the Mac OS section, it is *not* mentioned.

I checked my own home dir and only found folders from ~1 year old versions.

I’d say it’s safe to remove it.

vitorgalvao: Then we should also remove it from the other casks. If any of you want to tackle that, I’d be appreciated. Merging this now.

tjanson: I’ll put it on my todo.

Casks/appcode-eap.rb
```diff
@@ -9,23 +9,12 @@
...
-                # TODO: expand/glob for '~/Library/Preferences/jetbrains.appcode.*.plist',
```
tjanson: While we’re at it though, what about this line (and lines like it)? They still make sense, right, because the globbing feature is planned?

vitorgalvao: Yes.

PR #2098, opened 2016-05-26
LibreOffice 5.2 alpha1

yurikoles:
Should I submit it under different name?


fanquake:
Maybe we can rename it to libreoffice-dev so it can include alpha & beta versions?

yurikoles:
@fanquake so let use -dev for beta and alpha and -rc for corresponding releases.

fanquake:
Are you also changing the actual cask name to libreoffice-dev ? I think we could use it for alpha, beta and rc.

yurikoles:
up & rebase

PR #2143, opened 2016-06-01
Adds omnigraffle-5 (specific version)

derphilipp:
Adds a explicit version for omnigraffle version 5 (instead of 6 or newer);


Casks/omnigraffle-5.rb
```diff
@@ -0,0 +1,13 @@
+cask 'omnigraffle-5' do
```
vitorgalvao: There shouldn’t be a `-` before the number.

derphilipp:
Thank you very much @adidalal and @vitorgalvao .

PR #2158, opened 2016-06-04
Update airmail-beta to version 3.0,369

mauricerkelly:

### DIFF
--- a/Casks/airmail-beta.rb
+++ b/Casks/airmail-beta.rb
@@ -1,11 +1,11 @@
 cask 'airmail-beta' do
-  version '3.0,368'
-  sha256 'b1da1a93f570c78e3507b75518b18d4f010fcd915b3be7477b0afc460801ae63'
+  version '3.0.0.369,248'
+  sha256 '5f93aa3f6c5f34a31002a2071c8b4f8948958cea4236a91ca97aeec46f98a037'
 
   # hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04 was verified as official when first introduced to the cask
-  url 'https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04/app_versions/247?format=zip&'
+  url "https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04/app_versions/#{version.after_comma}?format=zip&"
   appcast 'https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04',
-          checkpoint: '8a87130261608652a570838658fe20503d580407c704fbcb885f49d661e290e1'
+          checkpoint: 'dc31be4a6b08bfd479186e6aff447be49d0e76f41bd82c0ae150d75bd8f49dff'
   name 'AirMail'
   homepage 'http://airmailapp.com/beta/'
   license :commercial


Casks/airmail-beta.rb
```diff
@@ -1,11 +1,11 @@
 cask 'airmail-beta' do
-  version '3.0,368'
-  sha256 'b1da1a93f570c78e3507b75518b18d4f010fcd915b3be7477b0afc460801ae63'
+  version '3.0,369'
```
vitorgalvao: `version '3.0,369'`

Casks/airmail-beta.rb
```diff
@@ -1,11 +1,11 @@
...
+  sha256 '5f93aa3f6c5f34a31002a2071c8b4f8948958cea4236a91ca97aeec46f98a037'
 
   # hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04 was verified as official when first introduced to the cask
-  url 'https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04/app_versions/247?format=zip&'
+  url 'https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04/app_versions/248?format=zip&'
```
vitorgalvao: `url "https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04/app_versions/#{version.after_comma}?format=zip&"`

mauricerkelly: Maybe I'm missing something here, but `#{version.after_comma}` will give `369`, not the required `248`.

Casks/airmail-beta.rb
mauricerkelly: Is that not the same as what I have in the diff?

vitorgalvao: Oops, yes. I meant `version '3.0,248'`.

Casks/airmail-beta.rb
vitorgalvao: I meant https://github.com/caskroom/homebrew-versions/pull/2158/files/8d05791ad2139b05dcf419d89c21349eb5df1aac#r65803690.

Casks/airmail-beta.rb
mauricerkelly: That's not the build number that they list on their HockeyApp download page though: https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04/app_versions/248?format=zip&avtoken=6e9067ed84b90f5e15b6789e8b2a94e8b41cb689

`248` is the HockeyApp download number (and is incremental). `369` is the build number of the app itself. These don't map directly as there can be app build numbers that are not submitted to HockeyApp for distribution.

vitorgalvao: Seeing [the appcast](https://rink.hockeyapp.net/api/2/apps/84be85c3331ee1d222fd7f0b59e41b04), they seem to increment together.

> These don't map directly as there can be app build numbers that are not submitted to HockeyApp for distribution

We wouldn’t be able to get to those anyway, then, since the `url` is from `hockeyapp`.

mauricerkelly: They may increment together now, but they're not the same. Unless you want to put in some sort of constant to convert the app build number to the HockeyApp download number, (which I think is a bad idea as it is arbitrary and subject to breakage if they break the pattern), why can't it just be left as it has been for numerous iterations of this cask?

vitorgalvao: Because it makes updates easier.

We never cared about app versions. We care about `url` versions we an interpolate.

If the build number is important to you, then make it `version '3.0.369,278'`. The `.*_comma` version methods exist exactly for this.

mauricerkelly: I would say that the build number is very important (and not to me, but to most consumers of this cask). This app regularly releases multiple builds against a single app version (e.g. 3.0), and the build number is important so that it can be seen if the cask up-to-date in comparison to the build number that the user is currently running (the build number is identified on the app's About box).

I will update this version to be `3.0.0.369,248`.

mauricerkelly:
@vitorgalvao updated after PR feedback.

PR #2203, opened 2016-06-12
remove utorrent-beta

fanquake:
The mac beta hasn't been updated since 2014. See [here](http://macreleasenotes.utorrent.com/)


vitorgalvao:
Nice findings, with these outdated betas.

PR #2218, opened 2016-06-19
Delete sublime-text3.rb

vitorgalvao:
Now the default in the main repo.


jawshooah:
Should probably add `sublime-text2` here in its place, since 3 is still (somehow) officially in beta.

vitorgalvao:
Didn’t add `2` on purpose, to see if someone misses it. If they do, readding is easy.

PR #2229, opened 2016-06-22
Remove PokerStars EU

fanquake:
The download link now just redirects to the homepage.


vitorgalvao:
It’s supposed to. This is very much alive. Where are you based at? Pokerstars is awful at redirecting, but they always redirect you depending on which country you are. They’re particularly awful because they redirect you to the home page of your country instead of the equivalent of the link you clicked on.

PR #2299, opened 2016-07-05
Delete iterm2-legacy.rb

vitorgalvao:
As per [the rules](https://github.com/caskroom/homebrew-versions#acceptable-casks):

Pinging @adidalal @hakamadare.


adidalal:
AppleScript compatibility breaks - I upgraded the few things I had, but it seems like a large enough change to warrant the -legacy.

Fine with the removal personally, though.

vitorgalvao:
> AppleScript compatibility breaks

Sounds like a “clear demonstrable need” to me. Closing. We can remove this later, after the one year period.

PR #2398, opened 2016-08-04
Cask repair: vlc-nightly

MarkGriffiths:
#### Editing an existing cask

- [x] `brew cask audit --download vlc-nightly` is error-free.
- [x] `brew cask style --fix vlc-nightly` left no offences.

I was getting the error:

```shell
Error: It seems the App source is not there: '/usr/local/Caskroom/vlc-nightly/latest/vlc-3.0.0-git/VLC.app'
```

- Added correct container type 'dmg'
- Added target to allow installation alongside current VLC.app, sharing prefs.
- Added two caveats(!)

Up until yesterday I was manually installing (_I should start using casks more!_) and added the caveats as both have affected me while hacking on a libass-based tool that needs VLC 3.0.0 nightlies.

Fixing prefs corruption is as simple as clicking the 'Reset all' button in VLC's prefs panel, but if someone has spent time tuning up their config in VLC 2.x, it's probably worth warning that sharing prefs with the nightly could cause loss of that configuration.

### DIFF
--- a/Casks/vlc-nightly.rb
+++ b/Casks/vlc-nightly.rb
@@ -10,9 +10,5 @@ cask 'vlc-nightly' do
   depends_on macos: '>= :lion'
   container type: :dmg
 
-  # Renamed to prevent conflict with current VLC, if installed.
-  app 'VLC.app', target: 'VLC (Nightly).app'
-
-  caveats "As prerelease software, #{token} could be unstable, defective or not run at all."
-  caveats "While #{token} will install alonside a current VLC, they share preferences which could become corrupt and require resetting."
+  app 'VLC.app'
 end
--- a/Casks/vlc-nightly.rb
+++ b/Casks/vlc-nightly.rb
@@ -8,7 +8,6 @@ cask 'vlc-nightly' do
   license :oss
 
   depends_on macos: '>= :lion'
-  container type: :dmg
 
   app 'VLC.app'
 end
--- a/Casks/vlc-nightly.rb
+++ b/Casks/vlc-nightly.rb
@@ -8,6 +8,11 @@ cask 'vlc-nightly' do
   license :oss
 
   depends_on macos: '>= :lion'
+  container type: :dmg
 
-  app 'vlc-3.0.0-git/VLC.app'
+  # Renamed to prevent conflict with current VLC, if installed.
+  app 'VLC.app', target: 'VLC (Nightly).app'
+
+  caveats "As prerelease software, #{token} could be unstable, defective or not run at all."
+  caveats "While #{token} will install alonside a current VLC, they share preferences which could become corrupt and require resetting."
 end


Casks/vlc-nightly.rb
```diff
@@ -8,6 +8,11 @@
   license :oss
 
   depends_on macos: '>= :lion'
+  container type: :dmg
 
-  app 'vlc-3.0.0-git/VLC.app'
+  # Renamed to prevent conflict with current VLC, if installed.
+  app 'VLC.app', target: 'VLC (Nightly).app'
+
+  caveats "As prerelease software, #{token} could be unstable, defective or not run at all."
```
vitorgalvao: Please remove the `caveats`. They need to pertain to installation only.

Casks/vlc-nightly.rb
```diff
@@ -8,6 +8,11 @@
   license :oss
 
   depends_on macos: '>= :lion'
+  container type: :dmg
 
-  app 'vlc-3.0.0-git/VLC.app'
+  # Renamed to prevent conflict with current VLC, if installed.
```
vitorgalvao: Do not rename. We do not make those alterations.

vitorgalvao:
For more details, see [this](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases), [this](https://github.com/caskroom/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md), and [this](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/caveats.md).

MarkGriffiths:
Thanks, that's fair enough… I've only just noticed that brew-cask won't overwrite an existing .app, even if it's from another formula, which was my main concern as the nightly is still a bit twitchy.

Casks/vlc-nightly.rb
```diff
@@ -8,6 +8,7 @@
   license :oss
 
   depends_on macos: '>= :lion'
+  container type: :dmg
```
vitorgalvao: Are you sure this is needed?

MarkGriffiths: Very much so. The source URL doesn't specify it's a DMG in any way which is why I was getting the error:

```
Error: It seems the App source is not there: '/usr/local/Caskroom/vlc-nightly/latest/vlc-3.0.0-git/VLC.app'
```

MarkGriffiths: And you might have missed my edited response above… is that correct?

vitorgalvao: That error is unrelated. It means the directory structure inside the package changed, not that it cannot detect the package type. It’s unlikely this line is needed.

As for different apps, whatever you like as the solution is fine, we just don’t impose namings without good cause.

MarkGriffiths: It fails without that line with both the new or old app declaration.

vitorgalvao: Just tried it. As expected, it works fine without the line.

vitorgalvao:
Thank you for the contribution. It needed some fixes, so they were made in 57c5fe12b5048e3d42b1ecd367bfe091a22addd1. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

MarkGriffiths:
Thanks… I was certain it wasn't working without that, but probably had got something out of sequence, as far as I can tell the structure hasn't changed - I've been manually downloading the same dmg for some time.

As I was testing I  noticed that `brew-cask uninstall` will remove the app belonging to the other formula if --appdir isn't specified.

So… 
- Installing cask `vlc-nightly` with `--appdir=~/Applications` places `VLC.app` correctly.
- Installing cask `vlc` places `VLC.app` into `/Applications`, also correctly.
- Then running `brew-cask uninstall vlc-nightly` removes the `VLC.app` in `/Applications`, incorrectly.

This is exactly what I trying to avoid by adding the target originally and as far as I can tell there are no alternatives to `--appdir` for installing different versions of the same app from different formula. Is this by design - so I need to remember to always add the --appdir, or sub-optimally swap the location of vlc-current and vlc-nightly? Is it something I should open an issue for?

vitorgalvao:
This is by design. We have some ideas for how to offset this (like changing `target` at install time), but nothing done yet.

MarkGriffiths:
Thanks… one last question if I may… would the best solution actually be to set up my own personal tap to handle it just the way I need? Does this cause any conflict?

vitorgalvao:
No conflict. To me, the best part of the homebrew system is how easy it is to set your own tap.

MarkGriffiths:
Thats brilliant… I couldn't live with it now and the more I can use it for, the better. Thanks for all your work looking after it 👍

PR #2419, opened 2016-08-08
Remove caveat about losing data in torbrowser-alpha

olbrew:

Since version 6 the Torbrowser is codesigned on MacOS. Due to the new folder layout necessary for this change the user profile is not stored in the application bundle anymore. This means that you will not lose your settings anymore when overwriting an existing Torbrowser. The caveat can thus be removed.


jawshooah:
Thanks!

PR #2525, opened 2016-09-01
Remove uninstall delete directive from IntelliJ IDEA casks

metyl:
### Checklist

None of Intellij IDEA product is linked into the `/usr/local/bin` during installation so they should be removed during uninstallation.

### DIFF
--- a/Casks/intellij-idea-ce-eap.rb
+++ b/Casks/intellij-idea-ce-eap.rb
@@ -12,8 +12,6 @@ cask 'intellij-idea-ce-eap' do
 
   app "IntelliJ IDEA #{version.before_comma} CE EAP.app"
 
-  uninstall delete: '/usr/local/bin/idea'
-
   zap delete: [
                 "~/Library/Application Support/IdeaIC#{version.major_minor}",
                 "~/Library/Preferences/IdeaIC#{version.major_minor}",
--- a/Casks/intellij-idea-ce.rb
+++ b/Casks/intellij-idea-ce.rb
@@ -12,8 +12,6 @@ cask 'intellij-idea-ce' do
 
   app 'IntelliJ IDEA CE.app'
 
-  uninstall delete: '/usr/local/bin/idea'
-
   zap delete: [
                 "~/Library/Application Support/IdeaIC#{version.major_minor}",
                 "~/Library/Preferences/IdeaIC#{version.major_minor}",
--- a/Casks/intellij-idea-eap.rb
+++ b/Casks/intellij-idea-eap.rb
@@ -11,8 +11,6 @@ cask 'intellij-idea-eap' do
 
   app "IntelliJ IDEA #{version.before_comma} EAP.app"
 
-  uninstall delete: '/usr/local/bin/idea'
-
   zap delete: [
                 "~/Library/Caches/IntelliJIdea#{version.major_minor}",
                 "~/Library/Logs/IntelliJIdea#{version.major_minor}",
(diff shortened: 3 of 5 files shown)


reitermarkus:
Move them to `zap` instead, then, rather than completely deleting them.

metyl:
These links were never created by any of these casks, then You try to create them using
 
> binary "#{appdir}/IntelliJ IDEA.app/Contents/MacOS/idea", target: 'idea' 

the `idea` link will be broken and app will not start. You will get 

> 2016-09-01 11:23:11.565 idea[58937:1245253] allVms required (null)
2016-09-01 11:23:11.570 idea[58937:1245253] Info.plist is corrupted, Absent JVMOptions key.

Moreover they are in conflict with each other because each of them is the same so potentially  uninstalling `intellij-idea-eap` will remove link from `intellij-idea-ce`.

That's why IMO it is better to delete them completely.

jawshooah:
I'm not sure I understand: if, as you say, it's better to delete `/usr/local/bin/idea`, why would we remove that line from `uninstall`?

metyl:
It is better to remove this line `uninstall delete: '/usr/local/bin/idea'` from all casks because: 

- none of this casks have ever create this link
- if we try to create it, it will not work
- each cask is removing the same link '/usr/local/bin/idea' so they are interfering with each other

jawshooah:
Looks like these `uninstall` stanzas were added by @Jerzerak. The `idea` CLI also [appears to be official](https://www.jetbrains.com/help/idea/2016.2/working-with-intellij-idea-features-from-command-line.html#d946334e17), and may be automatically added through some setting in the IDE. @Jerzerak, is that the case? Does IntelliJ automatically link/copy the `idea` binary into `/usr/local/bin`?

metyl:
IMO @Jerzerak added them by copy-and-paste.
The CLI is supported but `idea` does **not** automatically link/copy the idea binary into `/usr/local/bin`, there is some problem with luncher.

To test it, try to link `idea` manually by `ln -s /Applications/IntelliJ\ IDEA.app/Contents/MacOS/idea /usr/local/bin/` and try to run it by `idea`, it will fail.

jawshooah:
Funnily enough, I just found a `/usr/local/bin/idea` on my machine, and it wasn't a symlink. I've created a gist of it here: https://gist.github.com/jawshooah/0f24fabcc280c702a61be471557ebb10

I never manually installed this, so it must have been dropped there by the IDE. Let's leave the `uninstall` as it is.

Jerzerak:
@metyl @jawshooah I added that line to the `uninstall` stanza after monitoring what files were created upon installation and launch of several different IntelliJ IDEA versions and flavors (off the top of my head it was at least EAP, CE, & Ultimate). If we remove that line the applications eventually become confused as to where their config files are located and what version/flavor of the application they should be.

From what I could gather the file is created upon launch, if it is not already present, so there shouldn't be any downside to deleting it upon uninstallation as any other present IDEA versions/flavors should create a new `/usr/local/bin/idea` file automatically. I haven't tested this since I added the line so things might've changed but as long as things are working as intended I'd leave it be.

jawshooah:
Thanks for the info, @Jerzerak. Looks like leaving the `uninstall` stanza alone is the right move.

metyl:
I've check with current version of IntelliJ IDEA (2016.2.3) and latest EAP (2016.3 build 163.3512.7) and none of them is creating/copying anything as `/usr/local/bin/idea`. Can You guys check it? @jawshooah Yours `/usr/local/bin/idea` is for 2016.1.1 version with old location of `--caskroom` maybe this is the problem?

metyl:
I've figured it out. You have to create this link **manually** in IntelliJ IDEA by `Tools->Create Command-Line launcher...` and specify the name of the script and the path where it should be created.

PR #2555, opened 2016-09-08
Add British English version of Adobe Photoshop

danielbayley:
### Checklist
- [x] `brew cask audit --download adobe-photoshop-cc-gb` is error-free.
- [x] `brew cask style --fix adobe-photoshop-cc-gb` left no offenses.
- [x] Checked there are no open [pull requests](https://github.com/caskroom/homebrew-versions/pulls) for the same cask.
- [x] Checked there are no closed [issues](https://github.com/caskroom/homebrew-versions/issues) where that cask was already refused.
- [x] When naming the cask, followed the [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] `brew cask install adobe-photoshop-cc-gb` worked successfully.
- [x] `brew cask uninstall adobe-photoshop-cc-gb` worked successfully.

### DIFF
--- a/Casks/adobe-photoshop-cc-gb.rb
+++ b/Casks/adobe-photoshop-cc-gb.rb
@@ -1,0 +1,40 @@
+cask 'adobe-photoshop-cc-gb' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://trials3.adobe.com/AdobeProducts/PHSP/16/osx10/Photoshop_16_LS20.dmg',
+      user_agent: :fake,
+      cookies:    { 'MM_TRIALS' => '1234' }
+  name 'Adobe Photoshop CC 2015'
+  homepage 'https://adobe.com/products/photoshop'
+  license :commercial
+
+  conflicts_with cask: 'adobe-photoshop-cc'
+
+  preflight do
+    system '/usr/bin/sudo', '-E', '--', "#{staged_path}/Adobe Photoshop CC 2015/Install.app/Contents/MacOS/Install", '--mode=silent', "--deploymentFile=#{staged_path}/Adobe\ Photoshop\ CC\ 2015/Deployment/en_GB_Deployment.xml"
+  end
+
+  uninstall_preflight do
+    uninstall_xml = "#{staged_path}/uninstall.xml"
+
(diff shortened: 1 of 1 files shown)


reitermarkus:
@vitorgalvao, I know [we've had this discussion before](https://github.com/caskroom/homebrew-cask/pull/14367/files), but I would still argue that the default casks should install the localized version matching the system language as this is also what happens when you manually install it, and only have `-en` casks for people that, as you said

> prefer their system in one language (many times their native one) and their apps in another (mostly english)

vitorgalvao:
@reitermarkus But can we be consistent with that rule across casks? Because I think we cannot, and that seems like a bad experience to me (users will be surprised when some localize and the others not, even the ones that should).

I wouldn’t oppose at all a solution like `--localized`, as in `brew cask install --localized adobe-photoshop` (naturally, `--localized` could also be given as a `HOMEBREW_CASK_OPTS` option). That would also standardise our naming scheme for languages (we’d use the exact same as macOS).

On a tangential note, I’d prefer for UK English to be the default, but US English makes more sense as the default for most apps.

reitermarkus:
While the idea of a `--localized` option sounds nice, it would really only be usable for edge cases, and then you would still have to translate the specified language slug into the correct format, like here with `-` and `_`.

For `.app` casks and any other *native* applications, the translations are already bundled.

> But can we be consistent with that rule across casks? Because I think we cannot

I don't see why not, also I don't know if there even are other casks than the Adobe ones that are multilingual but aren't native apps.

danielbayley:
> I don't know if there even are other casks than the Adobe ones that are multilingual but aren't native apps.

Trust Adobe to screw it up…

reitermarkus:
Ok, so there are Firefox and Thunderbird, which for some reason use language packs instead of bundling translations natively, other than that, I don't think there are any.

vitorgalvao:
@reitermarkus Open an discussion issue on the main repo. If there are no objections, I’m fine with changing the rule.

vitorgalvao:
Closing, since merging languages is in progress and seems to have no objections.

PR #2558, opened 2016-09-09
Cleanup localized casks.

reitermarkus:


yurikoles:
shipit!

reitermarkus:
This depends on https://github.com/Homebrew/brew/pull/906, which is still in the works, so this cannot be shipped as of yet.

RobberPhex:
Homebrew/brew#906 was merged, so I think this pr can merge.

reitermarkus:
There is a bug in `brew cask audit`, which would result in this PR going green even if *some* of the languages failed. So this is on hold until https://github.com/Homebrew/brew/pull/1216 is merged and tagged.

PR #2587, opened 2016-09-15
Update thebrain9 cask to version 9.0.107.0

enogrob:
### Checklist

- [x] `brew cask audit --download thebrain9` is error-free.
- [x] `brew cask style --fix thebrain9.rb` left no offenses.

### DIFF
--- a/Casks/thebrain9.rb
+++ b/Casks/thebrain9.rb
@@ -1,6 +1,6 @@
 cask 'thebrain9' do
-  version '9.0.101.0'
-  sha256 '01c35e65b24e0de65b11a55148d68c3e1eb7fbec66934f49eae6bb6da30e88dc'
+  version '9.0.107.0'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
You are not adding a new cask. That section is not to fill.

Could you please tell us what is so unclear about that part of the instructions? You’re not the first person to fill it when they’re not supposed to, and I genuinely have no idea why.

enogrob:
...maybe because there is not an explicit instruction in case of a just a change version. At a certain point you are told to follow when you are adding a new cask.

vitorgalvao:
> there is not an explicit instruction in case of a just a change version

It’s the top of the checklist. The bottom tells you to fill it when **adding a new cask** (it’s even bolded). I still do not understand the confusion.

This is an honest question, I do not get while some people fill that section.

PR #2721, opened 2016-10-06
Remove license from all casks

abookyun:
> @reitermarkus: Every cask has a license stanza, which is, and never was, used for anything.

So, I remove all license stanza from all casks and documentation.

discussion from here: https://github.com/caskroom/homebrew-cask/issues/16692 and https://github.com/caskroom/homebrew-cask/issues/17437

### DIFF
--- a/Casks/1password-beta.rb
+++ b/Casks/1password-beta.rb
@@ -5,7 +5,6 @@ cask '1password-beta' do
   url "https://cache.agilebits.com/dist/1P/mac4/1Password-#{version}.zip"
   name '1Password'
   homepage 'https://agilebits.com/downloads'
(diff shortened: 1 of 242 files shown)


vitorgalvao:
Not sure this PR should exist just yet. Reason being that since Homebrew now obeys tags, we actually need a new release after your change is merged before we can actually do this. With that in mind, these PRs will just be on hold and there will likely be conflicts when we can actually merge them.

These are pretty easy to make, so we can hold off a bit. I see you already submitted a PR to deprecate the stanza (Thank you!).

abookyun:
@vitorgalvao Sure, let's close this first. And glad that was helped. :)

PR #2804, opened 2016-10-23
Update slack-beta to 2.3.0

soupdiver:


Casks/slack-beta.rb
```diff
@@ -1,6 +1,6 @@
 cask 'slack-beta' do
-  version '2.2.6-beta1'
-  sha256 '7166abdf83772bcfc4ada304499dd6b1de0aafb4d9fdde0980a3a665e1d7c1d4'
+  version '2.3.0'
```
vitorgalvao: Is this still a beta? Doesn’t look like it.

soupdiver: It's the version you get via: https://slack.com/beta

PR #2829, opened 2016-10-30
Unison 2.40.69

sebastienbarre:
## Rationale

(also see discussion in caskroom/homebrew-cask#26206)

Unison is both a client and a server. Unfortunately, the choice of client is dictated by the server *minor version* -- client and server minor versions *have* to match on both sides. Luckily, there have been only so many "official" minor versions in the past 8 years -- 2.27, 2.32, 2.40, and 2.48.

Right now the official [unison cask](https://github.com/caskroom/homebrew-cask/blob/master/Casks/unison.rb) provides 2.48 (unless you have an old OSX). e.g.:

```
 if MacOS.version <= :mountain_lion
    version '2.40.69'
    sha256 '2bcc460511f2b43fa1613cc5f9ba4dd59bb12d40b5b9fb2e9f21adaf854bcf3b'
    url "http://unison-binaries.inria.fr/files/Unison-#{version}_x64.dmg"
  else
    version '2.48.15'
    sha256 '89894d14c9ff3c4d6195cb6a8065a2849e6ad55951799eedf8879e1a257d3e11'
    url "http://unison-binaries.inria.fr/files/Unison-OS-X-#{version}.zip"
```

But I need to install 2.40 on an new/recent OSX, because I talk to a server that runs 2.40. This `unison240` cask strips 2.48 from the official cast and only provides the latest stable 2.40.69 version, circa 2014 (see official [Unison Binaries](http://unison-binaries.inria.fr/)).

Thank you

## Boilerplate

- [x] Checked that the cask was not already refused in [closed issues].

### DIFF
--- a/Casks/unison240.rb
+++ b/Casks/unison240.rb
@@ -1,0 +1,13 @@
+cask 'unison240' do
+  version '2.40.69'
+  sha256 '2bcc460511f2b43fa1613cc5f9ba4dd59bb12d40b5b9fb2e9f21adaf854bcf3b'
+
+  # unison-binaries.inria.fr was verified as official when first introduced to the cask
+  url "http://unison-binaries.inria.fr/files/Unison-#{version}_x64.dmg"
+  name 'Panic Unison'
+  homepage 'https://www.cis.upenn.edu/~bcpierce/unison/'
+
+  depends_on arch: :x86_64
+
+  app 'Unison.app'
+end
--- a/Casks/unison240.rb
+++ b/Casks/unison240.rb
@@ -7,7 +7,5 @@ cask 'unison240' do
   name 'Unison'
   homepage 'https://www.cis.upenn.edu/~bcpierce/unison/'
 
-  depends_on arch: :x86_64
-
   app 'Unison.app'
 end
--- a/Casks/unison240.rb
+++ b/Casks/unison240.rb
@@ -4,7 +4,7 @@ cask 'unison240' do
 
   # unison-binaries.inria.fr was verified as official when first introduced to the cask
   url "http://unison-binaries.inria.fr/files/Unison-#{version}_x64.dmg"
-  name 'Panic Unison'
+  name 'Unison'
   homepage 'https://www.cis.upenn.edu/~bcpierce/unison/'
 
   depends_on arch: :x86_64


Casks/unison240.rb
```diff
@@ -0,0 +1,13 @@
+cask 'unison240' do
+  version '2.40.69'
+  sha256 '2bcc460511f2b43fa1613cc5f9ba4dd59bb12d40b5b9fb2e9f21adaf854bcf3b'
+
+  # unison-binaries.inria.fr was verified as official when first introduced to the cask
+  url "http://unison-binaries.inria.fr/files/Unison-#{version}_x64.dmg"
+  name 'Panic Unison'
```
reitermarkus: This is actually the non-Panic Unison. Would be great if you could also fix the `panic-unison` and `unison` cask in the main repo to match this.

sebastienbarre: I took the official `unison` cask and stripped only the 2.48 version. It advertises itself as "Panic Unison":
```
$ brew cask info unison
unison: 2.48.15
From: https://github.com/caskroom/homebrew-cask/blob/master/Casks/unison.rb
==> Name
Panic Unison
==> Artifacts
Unison.app (app)
```
The other, panic-unison, seems deprecated:
```
$ brew cask info panic-unison
panic-unison: 2.2
https://panic.com/unison/
From: https://github.com/caskroom/homebrew-cask/blob/master/Casks/panic-unison.rb
==> Name
Unison
==> Artifacts
Unison.app (app)
==> Caveats
panic-unison has been officially discontinued upstream.
```

I didn't know if "fixing" the name would break something else so I left it as it was in the official cask. Would you like me to change the name field only? I'm not sure what to do about `panic-unison`.

sebastienbarre: My gut feeling was that `homebrew-versions` should use the same name as the official cask, to avoid confusion. If somebody sees the official 2.48 as being named "Panic Unison", and decide to search for 2.40 in `homebrew-versions`, I assume they would expect the same name, even though it was incorrect in the first place (2.40 predates 2.48)?

sebastienbarre: Done

vitorgalvao:
This cask is a bad candidate for this repo and we should refuse it. You said it yourself *you* need this. You’re likely the only person that does.

For our [acceptable casks policy](https://github.com/caskroom/homebrew-versions#acceptable-casks):

> `caskroom/versions` is not intended to be used for all and any old versions you personally require for xyz project; casks submitted here should be expected to be used by a reasonable number of people and supported by contributors long-term.

> You may wish to consider hosting your own [tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) for casks you wish to personally support that do not meet the above standards.

Thank you for the contribution, but this is exactly the type of clutter we want to avoid.

sebastienbarre:
@vitorgalvao Could you please clarify? I can only speak for myself, yes -- one can assume anybody who submits a cask has a some interest in it, right? Why are there 4 versions of Adobe Lightroom for example? Who uses 2013's 4.4.1?

- the choice of Unison client version is dictated by the version of the server you are talking to -- this sounds like a good case for keeping an old, stable, official version,
- there are only so many *actual* versions of Unison, in the past 8 years the community has been only using 2.27, 2.32, 2.40 (around 2014), and 2.48.

sebastienbarre:
It would be nice if there was some consensus on that decision, from more than one member? @reitermarkus seemed OK with this PR. @victorpopkov @fanquake?

vitorgalvao:
> one can assume anybody who submits a cask has a some interest in it, right?

Depends on the user. Many submit just for completeness, which is especially true for new apps. The point here is if **anybody else** has interest in this.

> Why are there 4 versions of Adobe Lightroom for example?

I only see three, and one of them is actually a requirement for the main cask (blame Adobe). Either way, we still have a lot of casks remaining from the days before making those policies. Some still need cleaning.

We also have a way to deal with those. In those same rules:

> Legacy versions of gratis or open-source software that were accepted should be removed after one year.

I do see what you mean here, though, and I misread the need for this. I was also under the impression this was the other, abandoned, version, so it seems fine. Reopening.

I will, however, ask that you refrain from spamming random maintainers when a decision does not go your way. In all honesty, it gave a bad taste to me reverting this as I was writing it. We have rules for a reason, and those rules are discussed and meant to be followed **because they have a reason to exist**. Many times a maintainer (including myself) forgets a specific rule and is OK with it when reminded of it, because they all were discussed. If @reitermarkus though this to be a special case, he would have the time to reply and we would reach consensus. Closing issues is not the end of the discussion.

We are not your personal repository, and do not have to host every little thing you want. Our organisation is more important than a single user’s personal needs. Again, you can always make a personal tap and are in no way beholden to us.

reitermarkus:

I think the need for this was demonstrated quite clearly by @sebastienbarre.

However, I think we should tighten this policy, since this rule is really vague. And especially in combination with

> Legacy versions of gratis or open-source software that were accepted should be removed after one year.

this will only lead to frustration. I think we should prefer custom taps over adding casks which we will have to delete manually a year after.

vitorgalvao:
> I think the need for this was demonstrated quite clearly by @sebastienbarre.

It was. Like I said:

> I misread the need for this. I was also under the impression this was the other, abandoned, version, so it seems fine. Reopening.

My fault, there.

> I think we should prefer custom taps over adding casks which we will have to delete manually a year after.

Agreed.

reitermarkus:
> It was. Like I said:

Ah, yes, I overlooked that part of your comment.

sebastienbarre:
My apologies for roping more people in. This was no acceptable. I got unnecessarily frustrated when you wrote "we should refuse it" and closed the PR -- it looked like you were slamming the door.

Unison's version numbering is not... great, to say the list, but for all intents and purposes 2.48, 2.40, and 2.32 before are considered "major" versions, with incompatible archive formats -- the same way Lightroom 6, 5, and 4 have incompatible catalog formats. They have a long lifetime as well.

I'm committed to maintain this 2.40 cask. This stable version was last updated in December 2014, so I don't expect much changes or a bump in release number.

Let me know how you'd like to proceed and I'll follow the repo rules closely. 
Thank you

vitorgalvao:
All is well.

The cask seems fine, but you can actually remove the `depends_on arch: :x86_64` line since we realistically only support one arch (the macOS versions we support are all 64-bit Intel).

Other than that, looks fine to merge.

sebastienbarre:
Done.

PR #2890, opened 2016-11-18
Fix homepages in 22 casks

victorpopkov:

---
Mostly done to avoid unnecessary redirects. HTTPS in some cases is removed since it's not available anymore.

### DIFF
--- a/Casks/vmware-fusion6.rb
+++ b/Casks/vmware-fusion6.rb
@@ -4,7 +4,7 @@ cask 'vmware-fusion6' do
 
   url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-#{version}-light.dmg"
   name 'VMware Fusion'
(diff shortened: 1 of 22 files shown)


Casks/vmware-fusion6.rb
```diff
@@ -4,7 +4,7 @@
 
   url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-#{version}-light.dmg"
   name 'VMware Fusion'
-  homepage 'https://www.vmware.com/products/fusion/'
+  homepage 'http://www.vmware.com/products/fusion/'
```
fanquake: Either of those direct me to `https://www.vmware.com/products/fusion.html`, so I think the https should be left.

victorpopkov: Fixed. Thank you for noticing. 👍

PR #2911, opened 2016-11-22
Update sourcetree-beta to latest

GregaVrbancic:

### DIFF
--- a/Casks/sourcetree-beta.rb
+++ b/Casks/sourcetree-beta.rb
@@ -1,9 +1,9 @@
 cask 'sourcetree-beta' do
-  version '2.4b6'
-  sha256 '97e30fba2c7eb8e62d81934c284adfc69559512c637653a0485de6a402e6f575'
+  version :latest
+  sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you, but users overwhelmingly prefer versioned casks.

GregaVrbancic:
No problem, that is fine, but you better then remove current sourcetree-beta cask because the download URL does not exist anymore, so the cask is broken. Also there [is no official versioned beta release](https://bitbucket.org/atlassianlabs/sourcetree-betas/downloads?tab=downloads), just the latest one to which I updated this cask.

vitorgalvao:
In that case, that is fine as it is.

vitorgalvao:
Actually, just noticed this is indeed wrong, so I’ll revert it. The link you changed it to goes to the same file as the one that was already present. This has an appcast, it can continue to be versioned.

vitorgalvao:
And the download URL does indeed exist and works fine.

GregaVrbancic:
Great that the url is now working, yesterday certainly was not. Anyway, I'm glad that the cask should work now.

PR #2988, opened 2016-12-09
Created josm-latest.rb

Andygol:

- [x] Checked that the cask was not already refused in [closed issues].

for testing latest version of JOSM

### DIFF
--- a/Casks/josm-latest.rb
+++ b/Casks/josm-latest.rb
@@ -1,0 +1,11 @@
+cask 'josm-latest' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://josm.openstreetmap.de/jenkins/job/JOSM/jdk=JDK8/lastSuccessfulBuild/artifact/dist/josm-custom-macosx.zip'
+  name 'josm latest'
+  homepage 'https://josm.openstreetmap.de/'
+
+  # Renamed to avoid removing tested (stable) version of JOSM
+  app 'JOSM.app', target: 'JOSM-latest.app'
+end
--- a/Casks/josm-latest.rb
+++ b/Casks/josm-latest.rb
@@ -6,6 +6,5 @@ cask 'josm-latest' do
   name 'josm latest'
   homepage 'https://josm.openstreetmap.de/'
 
-  # Renamed to avoid removing tested (stable) version of JOSM
-  app 'JOSM.app', target: 'JOSM-latest.app'
+  app 'JOSM.app'
 end


vitorgalvao:
`lastest` is not something we do. Is this a development version, or simply the latest stable version?

Andygol:
@vitorgalvao 
here is more details about josm release system -  https://josm.openstreetmap.de/wiki/Releases 

>The JOSM editor is kept always in a releasable state.
>Each night a new software is build (called latest) which can be used.

Andygol:
I used https://github.com/caskroom/homebrew-cask/blob/master/Casks/soundnode.rb as example

Casks/josm-latest.rb
```diff
@@ -0,0 +1,10 @@
+cask 'josm-latest' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://josm.openstreetmap.de/jenkins/job/JOSM/jdk=JDK8/lastSuccessfulBuild/artifact/dist/josm-custom-macosx.zip'
+  name 'josm latest'
+  homepage 'https://josm.openstreetmap.de/'
+
+  app 'JOSM.app'
```
Andygol: It would be nice if we had the opportunity to use both versions of Josm (`tested` and `latest`) simultaneously. But unfortunately they both have `JOSM.app` inside their packages. To prevent conflict between versions I added `target: JOSM-latest.app` 

 https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases

>With app, use it when it makes sense functionally and document your reason clearly in the Cask: was it for clarity; for consistency; **to prevent conflicts**; …

@vitorgalvao could you revert your changes  to previous state?

Andygol: Approach with `josm` and `josm-latest` is present in Ubuntu

vitorgalvao: We don’t follow what other package managers do. We have our own rules.

Preventing conflicts is a different matter, it’s for different yet related apps with the same name. This is a different version of the same app. See https://github.com/caskroom/homebrew-cask/issues/13201. We care about making the installation as close as possible to the manual one, and arbitrarily changing the name of app bundles is not part of it. If the developers gave them the same name, they likely have no intention of you running both.

You’re free to change the name for yourself, but we won’t impose that preference on every user.

Andygol: Today I faced with the issue when I was trying to install **josm-latest** along with **josm.app**

```
mbp:~ user$ brew cask install josm-latest
==> Satisfying dependencies
==> Downloading https://josm.openstreetmap.de/jenkins/job/JOSM/jdk=JDK8/lastSuccessfulBuil
==> No checksum defined for Cask josm-latest, skipping verification
==> Installing Cask josm-latest
Error: It seems there is already an App at '/Applications/JOSM.app'.
```

So, the following to the approach to have separate apps for `josm` and `josm-lates` should solve this problem.
For now, I've just edited my local cask to have both `josm` and `josm-lates` on my Mac.

@vitorgalvao is it possible to turn back to the initially proposed version for this cask?

PR #2994, opened 2016-12-12
Update pycharm-eap.rb

miccal:
Fix interpolation in `app` stanza.


reitermarkus:
All the `zap`s should also be in double quotes.

miccal:
Thanks @reitermarkus, submitted.

PR #3066, opened 2016-12-27
Add eXist-db.app v3.0RC1

joewiz:

- [X] Named the cask according to the [token reference].
    **Note**: Much as @alyssais did in https://github.com/caskroom/homebrew-versions/pull/2578, I opted to use "rc" as the suffix rather than a version number (like 3). This is because this recipe is intended for the *current* release candidate (eXist's convention for beta/preview releases), regardless of the current major/minor version number.
- [X] `brew cask install {{cask_file}}` worked successfully.
- [X] `brew cask uninstall {{cask_file}}` worked successfully.
- [X] Checked there are no [open pull requests] for the same cask.
- [X] Checked that the cask was not already refused in [closed issues].

### DIFF
--- a/Casks/exist-db3.rb
+++ b/Casks/exist-db3.rb
@@ -1,4 +1,4 @@
-cask 'exist-db-rc' do
+cask 'exist-db3' do
   version '3.0.RC1'
   sha256 '086f0813e526047d9cebca5be8d5f7012bb5d2cdb14e8ae98f3297bba50b9a58'
 
--- a/Casks/exist-db-rc.rb
+++ b/Casks/exist-db-rc.rb
@@ -1,0 +1,11 @@
+cask 'exist-db-rc' do
+  version '3.0.RC1'
+  sha256 '086f0813e526047d9cebca5be8d5f7012bb5d2cdb14e8ae98f3297bba50b9a58'
+
+  # dl.bintray.com/existdb was verified as official when first introduced to the cask
+  url "https://dl.bintray.com/existdb/releases/eXist-db-#{version}.dmg"
+  name 'eXist-db'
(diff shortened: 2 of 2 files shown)


Casks/exist-db3.rb
```diff
@@ -0,0 +1,11 @@
+cask 'exist-db3' do
+  version '3.0.RC1'
+  sha256 '086f0813e526047d9cebca5be8d5f7012bb5d2cdb14e8ae98f3297bba50b9a58'
+
+  # dl.bintray.com/existdb was verified as official when first introduced to the cask
+  url "https://dl.bintray.com/existdb/releases/eXist-db-#{version}.dmg"
+  name 'eXist-db'
+  homepage 'http://exist-db.org/exist/apps/homepage/index.html'
```
vitorgalvao: We need the homepage for version 3.

joewiz:
Thank you for your review and merge! I am happy with exist-db3 as opposed to exist-db-rc. 

Two questions: 

1. When eXist-db v3.0 is released and I update the main cask to use v3.0 instead of v2.2, should I delete this cask, or keep it for people who explicitly want v3.x as opposed to "the current version" (i.e., even in the future after a hypothetical v4.x is released)? 

2. Since 2.2 still has a large install base, I imagine some users would still want that even once v3.0 is released. This suggests that adding a "eXist-db2" cask would make sense, until v2.x is considered obsolete, right?

vitorgalvao:
When `3` becomes stable, this cask should be removed.

One for version `2` should not be added. Later if it becomes necessary, someone else will do it.

joewiz:
Got it. Thanks for explaining; I think I'm starting to understand the approach.

PR #3067, opened 2016-12-27
Add eXist-db.app 3.0 (nightly build)

joewiz:

- [X] Named the cask according to the [token reference].
    **Note:** In using "nightly" suffix, I followed the nomenclature at https://github.com/caskroom/homebrew-versions/blob/master/README.md and followed the example of other nightly casks, e.g., https://github.com/caskroom/homebrew-versions/commit/d7ac7c4a1f6d6dd9938956b48309c0be5ebcffac.
- [X] `brew cask install {{cask_file}}` worked successfully.
    **Note:** In constructing the URL to the latest nightly, I followed the example of https://github.com/caskroom/homebrew-versions/blob/master/Casks/vlc-nightly.rb.
- [X] `brew cask uninstall {{cask_file}}` worked successfully.
- [X] Checked there are no [open pull requests] for the same cask.
- [X] Checked that the cask was not already refused in [closed issues].

### DIFF
--- a/Casks/exist-db-nightly.rb
+++ b/Casks/exist-db-nightly.rb
@@ -4,12 +4,12 @@ cask 'exist-db-nightly' do
 
   # static.adamretter.org.uk/exist-nightly was verified as official when first introduced to the cask
   url do
-    require 'nokogiri'
     require 'open-uri'
     base_url = 'http://static.adamretter.org.uk/exist-nightly'
     builds_url = "#{base_url}/table.html"
-    builds_doc = Nokogiri::HTML(open(builds_url))
-    latest_build_filename = builds_doc.xpath('//tr[td/a]').sort_by { |a| a.xpath('td[1]').text }.last.xpath('td[3]/a[3]/@href')
+    latest_build_filename = open(builds_url) do |io|
+      io.read.scan(%r{<tr>.*?<td>(.*?)</td>.*?<a href="([^\"]+)">dmg}m).max[1]
+    end
     "#{base_url}/#{latest_build_filename}"
   end
   name 'eXist-db'
--- a/Casks/exist-db-nightly.rb
+++ b/Casks/exist-db-nightly.rb
@@ -1,0 +1,19 @@
+cask 'exist-db-nightly' do
+  version :latest
+  sha256 :no_check
+
+  # static.adamretter.org.uk/exist-nightly was verified as official when first introduced to the cask
+  url do
+    require 'nokogiri'
+    require 'open-uri'
+    base_url = 'http://static.adamretter.org.uk/exist-nightly'
+    builds_url = "#{base_url}/table.html"
+    builds_doc = Nokogiri::HTML(open(builds_url))
+    latest_build_filename = builds_doc.xpath('//tr[td/a]').sort_by { |a| a.xpath('td[1]').text }.last.xpath('td[3]/a[3]/@href')
+    "#{base_url}/#{latest_build_filename}"
+  end
+  name 'eXist-db'
+  homepage 'http://exist-db.org/exist/apps/homepage/index.html'
+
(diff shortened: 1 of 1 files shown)


Casks/exist-db-nightly.rb
```diff
@@ -0,0 +1,19 @@
+cask 'exist-db-nightly' do
+  version :latest
+  sha256 :no_check
+
+  # static.adamretter.org.uk/exist-nightly was verified as official when first introduced to the cask
+  url do
```
vitorgalvao: That’s not really the purpose of `url do`, and users do prefer versioned builds (which this has) over `latest`.

That said, up-to-date nightlies with no `sha256` are more useful than out-of-date nightlies, so I don’t really have an issue with this *regarding nightlies only*.

But since this isn’t something we usually do, I’d like some further input from @caskroom/maintainers.

joewiz: @vitorgalvao Thank you for your review and for taking the time to explain this principle. If this PR is good for you and the other maintainers, accepting this PR would be a real help to our community. Thanks in advance.

adidalal: Should be OK, imo - could be an exception for the -versions repo only

adidalal: Not familiar with ruby - are those two requires inbuilt?

claui: > are those two requires inbuilt?

@adidalal Good catch! While `open-uri` [is built into the system Ruby](https://ruby-doc.org//stdlib-1.9.3/libdoc/open-uri/rdoc/OpenURI.html), `nokogiri` [appears](http://ruby.zigzo.com/2015/03/18/installing-nokogiri-on-yosemite) [not to](http://stackoverflow.com/questions/40038953/installing-nokogiri-on-mac-os-sierra-10-12).

The dependency on `nokogiri` needs to go; luckily, we’ve had ad-hoc workarounds in the past where [regular expressions did that particular job well](https://github.com/caskroom/homebrew-cask/blob/c389d9ccbb46d30b6ac1cbdbadf49591ca8ff6cd/Casks/audacity.rb#L5-L15).

<sup>(Yes, the `audacity` formula did break often, albeit for unrelated reasons. And no: regular expressions are [not a valid general-purpose solution](http://stackoverflow.com/a/1732454/17305).)</sup>

joewiz:
Okay, I'll adapt the cask to remove Nokogiri. I will work with the creator of the nightly site to pre-sort the entries so the newest is consistently on top.

claui:
@joewiz In case you want to work with what you have, something like

```
latest_build_filename = open(builds_url) do |io|
  io.scan(/<tr>.*?<td>(.*?)<\/td>.*?<a href="([^\"]+)">dmg/m).max[1]
end
```

might do the trick.

claui:
@joewiz If you run the example code, you’re going to discover it already takes care of the sorting.

`String#scan` (in its non-block form) returns an array of date-URL pairs.

To pick the pair with the most recent date, the `max` call is already sufficient. You can remove the `[1]` temporarily to inspect the result.

joewiz:
@claui Now I see! Thanks so much for this great suggestion! Revised PR to come shortly.

joewiz:
Okay, the revised PR is in, and CI and my own tests are passing.

claui:
Like @vitorgalvao already pointed out (and [anticipated](https://github.com/caskroom/homebrew-cask/pull/22014#issuecomment-226489708) back when we introduced the feature), I feel this particular use of `url do` is going to be really helpful for nightlies that don’t have a stable latest-version URL. I think this PR should be merged. Thanks @joewiz!

joewiz:
@claui @adidalal @vitorgalvao Great to hear that this PR proved to be a broader contribution. Thank you so much for all of your help!

PR #3190, opened 2017-01-24
Update and rename pdfpenpro.rb to pdfpen-pro.rb

miccal:
Renamed for consistency.


vitorgalvao:
The token reference is still paramount. `pro` is part of the name (and app bundle), and not separated. Consistency does not apply in this case.

PR #3191, opened 2017-01-24
Update and rename firefoxnightly.rb to firefox-nightly.rb

miccal:
Renamed for consistency.


vitorgalvao:
Same. Token reference.

PR #3246, opened 2017-01-31
Fix wmail-pre-release app name retargeting

fdrouet:

- [x] Checked the cask was not already refused in [closed issues].
- [x] Checked the cask is submitted to [the correct repo].

[the correct repo]: https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask

### DIFF
--- a/Casks/wmail-pre-release.rb
+++ b/Casks/wmail-pre-release.rb
@@ -11,5 +11,5 @@ cask 'wmail-pre-release' do
 
   conflicts_with cask: 'wmail'
 
-  app 'WMail pre-release.app'
+  app 'WMail.app', target: 'WMail pre-release.app'
 end


vitorgalvao:
[`target:` Should Only Be Used in Select Cases](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/app.md#target-should-only-be-used-in-select-cases).

This already has a `conflicts_with`, which is the whole point.

fdrouet:
I don't agree with your decision to close the PR because without this change the Cask is not technically able to install because the official app name in the dmg is `WMail.app` not `WMail pre-release.app`. But what I really want to have is the `WMail pre-release.app` app name to simply know that what is installed is the pre-release channel of the application.

For the `conflicts_with` parameter, I defined it because we can't install both casks (wmail and wmail-pre-release) because they will write their data at the same location on the disk.

vitorgalvao:
> But what I really want

And that is your personal desire and we don’t cater to individual user’s desires. Just like *you* want it to have a different name, a different user will want it to have *the original name the developer name gave it*. The latter case always takes precedence.

fdrouet:
Ok so please reopen the PR to let me fix the installation problem ;-)

vitorgalvao:
> Ok so please reopen the PR to let me fix the installation problem ;-)

[Done](https://github.com/caskroom/homebrew-versions/commit/2bd28e20b1f6813258e2805d3590bf45e17c8da0).

fdrouet:
ok got it.

PR #3290, opened 2017-02-09
Delete exist-db3

joewiz:
Now that 3.0 final has been released and is the current stable version of eXist-db (see the [PR updating the main exist-db cask](https://github.com/caskroom/homebrew-cask/pull/29978)), this versioned cask is no longer needed.


joewiz:
@vitorgalvao Thank you!

PR #3507, opened 2017-03-22
Remove outdated elcapitan version of safari-technology-preview

ebraminio:
Currently "version :latest" without turning sha256 to :no_check, won't work on an actual El Capitan, which is done by me, sorry (because previous value, 24, was also untrue indication of the version, as version 24 was not published on 20160926 obviously). Can we instead remove the platform specific conditioning and make things more maintainable?


miccal:
Thank you.

PR #3549, opened 2017-03-30
Update firefoxdeveloperedition to 54.0a2

0rax:

### DIFF
--- a/Casks/firefoxdeveloperedition.rb
+++ b/Casks/firefoxdeveloperedition.rb
@@ -1,34 +1,42 @@
 cask 'firefoxdeveloperedition' do
-  version :latest
-  sha256 :no_check
+  version '54.0a2'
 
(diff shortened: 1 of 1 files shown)


Casks/firefoxdeveloperedition.rb
```diff
@@ -1,5 +1,5 @@
 cask 'firefoxdeveloperedition' do
-  version :latest
+  version '54.0a2'
   sha256 :no_check
```
victorpopkov: Since now this cask includes version the `sha256` is also required.

0rax: I was wondering exactly how this can be done with a language stanza. I can only provide a sha256 for one of the installer or is it possible to set one for each language ?

0rax:
Updated to add `sha256` for each language

PR #3567, opened 2017-04-02
migrate wine-devel

yurikoles:
https://github.com/caskroom/homebrew-cask/pull/31769


vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/30079

PR #3588, opened 2017-04-05
Move back to latest version for Firefox Developer Edition

lucafavatella:
... as before ecc54919b.

This cask has recently been switched from latest to specific versions and hash, and since then has undergone several updates. I have just come to use it and I found the hash is already outdated:
```
==> Downloading https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-aurora/firefox-54.0a2.en-US.mac.dmg
==> Verifying checksum for Cask firefoxdeveloperedition
==> Note: running "brew update" may fix sha256 checksum errors
Error: sha256 mismatch
Expected: 910bde849efba3e0064997d357661ba6b71820533b095886521b770e30b09bbb
Actual: e0f4846fbfc02682eff27e318ef43eb6fab6168eba213e41b1382d8b9bc89b85
File: /Users/luca/Library/Caches/Homebrew/Cask/firefoxdeveloperedition--54.0a2.dmg
To retry an incomplete download, remove the file above.
```
Hence I propose to revert back to latest without hash.

----

### DIFF
--- a/Casks/firefoxdeveloperedition.rb
+++ b/Casks/firefoxdeveloperedition.rb
@@ -1,42 +1,34 @@
 cask 'firefoxdeveloperedition' do
-  version '54.0a2'
+  version :latest
+  sha256 :no_check
 
   language 'en', default: true do
-    sha256 '910bde849efba3e0064997d357661ba6b71820533b095886521b770e30b09bbb'
     'en-US'
   end
 
   language 'ja' do
-    sha256 'ae7a9f27cfb0e6e1aacdef07b6bec183247bc2263bc89e9b1fef266dc000910f'
     'ja-JP-mac'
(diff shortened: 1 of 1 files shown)


0rax:
My bad sorry, I made this update in #3549 as It was looking like the dev edition was versioned in the same way as a beta version. Though after taking a closer look at all the files, it seems like the Developer Edition is built daily without updating the version number (see https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-aurora/firefox-54.0a2.en-US.mac.json). It seems that there is no way to keep a specific version up and even if it is the case we would have to update the cask manually.

I am all for a revert, sorry for the inconvenience.

PR #3612, opened 2017-04-09
Update tap migrations - maltego

commitay:
I'm not sure if this should be deleted or renamed as `maltego-chlorine-ce` is now [`maltego-ce`](https://github.com/caskroom/homebrew-cask/blob/master/Casks/maltego-ce.rb)?


miccal:
Thank you.

PR #3703, opened 2017-04-21
Revert "Add Audacity Nightly"

miccal:
Reverts caskroom/homebrew-versions#3702

Audacity it uses [fosshub](https://www.fosshub.com/faq.html) as a distribution site, and we cannot have it in Cask.


sscotth:
Audacity nightlies are not on fosshub. They are on http://www.audacity.homerow.net and have been since around 2009.

miccal:
@sscotth nevertheless, the developers behind Audacity have requested that downloads not be made via things like Cask.

PR #3714, opened 2017-04-23
Bring back intellij-idea 2016.1

condemil:
After removal in https://github.com/caskroom/homebrew-versions/pull/3683

- [ ] `brew cask audit --download {{cask_file}}` is error-free.
- [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
- [ ] The commit message includes the cask’s name and version.

- [ ] Named the cask according to the [token reference].
- [ ] `brew cask install {{cask_file}}` worked successfully.
- [ ] `brew cask uninstall {{cask_file}}` worked successfully.
- [ ] Checked there are no [open pull requests] for the same cask.
- [ ] Checked the cask was not already refused in [closed issues].
- [ ] Checked the cask is submitted to [the correct repo].

[the correct repo]: https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask

### DIFF
--- a/Casks/intellij-idea20161.rb
+++ b/Casks/intellij-idea20161.rb
@@ -1,0 +1,25 @@
+cask 'intellij-idea20161' do
+  version '2016.1.4,145.2070'
+  sha256 'c6ed2455cfeeb542c646e6a3d639d32e85ebe8ead87a0f3744dfa3c0a920e2f9'
+
+  url "https://download.jetbrains.com/idea/ideaIU-#{version.before_comma}.dmg"
+  name 'IntelliJ IDEA Ultimate'
+  homepage 'https://www.jetbrains.com/idea/'
+
+  auto_updates true
+
+  app 'IntelliJ IDEA.app'
+
+  uninstall_postflight do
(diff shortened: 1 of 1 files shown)


leipert:
@condemil / @vitorgalvao: Is it really necessary to bring the cask back?

The only reason to bring the cask back would be for people that do not have a license for newer version. But by that logic, we would need to bring back other old jetbrains cask as well.

There are now 24 stable intellij versions alone and 112 stable versions of jetbrains products.
Maybe it would make more sense to move them to a new tap ? I Could also update the @jcbot to maintain such a tap.

condemil:
There was no any other non-eap JetBrains products removed by https://github.com/caskroom/homebrew-versions/pull/3683

Fallback license holders can use latest z version from x.y.z version format, as I can see [here](https://www.jetbrains.com/idea/download/previous.html) there are only 12 x.y intellij versions released since 2012.

vitorgalvao:
This is well covered in the [acceptable casks policy](https://github.com/caskroom/homebrew-versions#acceptable-casks). For at least one year, this cask has a place here.

> But by that logic, we would need to bring back other old jetbrains cask as well.

No one asked for those, so no. This one was specifically asked for. This is the correct repo.

Also, this is way different from the EAP casks, as this one will presumably never be updated, hence not making it a maintainability burden.

PR #3914, opened 2017-05-26
Remove java7

commitay:
Download has been moved behind a login.

Closes https://github.com/caskroom/homebrew-versions/issues/3902


miccal:
Pint @vitorgalvao - should we remove [java6](https://github.com/caskroom/homebrew-versions/blob/master/Casks/java6.rb) also?

commitay:
> should we remove java6 also?

> https://support.apple.com/kb/DL1572?locale=en_GB
> This package is exclusively intended for support of legacy software and installs the same deprecated version of Java 6 included in the 2014-001 and 2013-005 releases.

Two casks have a `java('6')` only dependency.
- [argouml](https://github.com/caskroom/homebrew-cask/blob/master/Casks/argouml.rb)
- [logitech-harmony](https://github.com/caskroom/homebrew-drivers/blob/master/Casks/logitech-harmony.rb)

I'll check and see if they work with java 8.

miccal:
@commitay never mind - we should definitely _not_ remove java6.

commitay:
> we should definitely not remove java6

Just out of curiosity, why?

commitay:
I've only been able to find one cask with a hard `java7` dependency and it requires JVM 1.7  or higher.

https://github.com/caskroom/homebrew-cask/pull/34960

to-kra:
whats is the solution for this ? only manual download from oracle page ?

vitorgalvao:
@to-kra Or add it to your personal tap.

abigailbunyan:
Another solution is to use the [`zulu7`](https://github.com/caskroom/homebrew-versions/blob/master/Casks/zulu7.rb) cask. Zulu is a certified build of OpenJDK produced by Azul Systems that should be around for a long time (they even offer JDK6 builds still).

SoylentGraham:
How do I use the zulu7 cask? I can't find any documentation on using (user-submitted?) casks(? taps?)

PR #3939, opened 2017-06-02
Delete firefoxdeveloperedition.rb

vitorgalvao:
Looks like it’ll be discontinued.


Nainterceptor:
Hi @vitorgalvao, according to this article : https://hacks.mozilla.org/2017/04/simplifying-firefox-release-channels/, firefox developer edition isn't discontinued, but switch stability from aurora channel to beta channel. fde will be less early but more stable. Can you revert this commit please ?

PR #4111, opened 2017-07-05
Fix style.

reitermarkus:


commitay:
> == Casks/adobe-photoshop-cs6.rb ==
C: 32: 21: Use %w or %W for an array of words.

I'm not sure what to do with this.

reitermarkus:
Travis might have `rubocop-cask` version 0.13.0 cached, so it won't update to 0.13.1.

PR #4190, opened 2017-07-22
add transmit 4

joshmanders:
Recently Panic released Transmit 5, this brings in Transmit 4 for those of us who don't want to upgrade.


victorpopkov:
Thank you for your contribution!

PR #4202, opened 2017-07-25
Update 1password-beta to 6.8

giannitm:

Additionally, if **updating a cask**:

- [ ] [If the `sha256` changed but the `version` didn’t](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256),
      provide public confirmation by the developer: {{link}}


giannitm:
This isn't actually a beta release of 1password, but it is the latest release available as of now. Should this cask be updated?

victorpopkov:
I suppose there is no point to update this cask to latest stable version.

If it's the same as the one published in the main [caskroom/homebrew-cask](https://github.com/caskroom/homebrew-cask) repository, there would be no point in this cask at all. I think it should be left as is until a new beta release is published.

PR #4332, opened 2017-08-26
Delete adium-beta.rb

vitorgalvao:
https://github.com/caskroom/homebrew-cask/pull/38016


vitorgalvao:
@reitermarkus Isn’t this version vulnerable as well? From the linked article in the other post, I see nothing saying it‘s fixed in the beta.

PR #4586, opened 2017-10-13
Remove adobe-digital-editions2

commitay:
https://github.com/caskroom/homebrew-versions/pull/1231

> https://www.adobe.com/support/digitaleditions/downloads.html
> Adobe Digital Editions 2.0.1
> 07/15/2013


ickc:
@commitay, why is this removed? I tried the link and it is still available? Isn't it the purpose of homebrew-versions to keep older versions around?

PR #4633, opened 2017-10-21
Remove jce-unlimited-strength-policy8

commitay:
`jce-unlimited-strength-policy8` doesn't work with the current `java8` Cask.

> http://www.oracle.com/technetwork/java/javase/8u152-relnotes-3850503.html

>  New Security property to control crypto policy
This release introduces a new feature whereby the JCE jurisdiction policy files used by the JDK can be controlled via a new Security property. In older releases, JCE jurisdiction files had to be downloaded and installed separately to allow unlimited cryptography to be used by the JDK. The download and install steps are no longer necessary. To enable unlimited cryptography, one can use the new crypto.policy Security property. If the new Security property (crypto.policy) is set in the java.security file, or has been set dynamically using the Security.setProperty() call before the JCE framework has been initialized, that setting will be honored. By default, the property will be undefined. If the property is undefined and the legacy JCE jurisdiction files don't exist in the legacy lib/security directory, then the default cryptographic level will remain at 'limited'. To configure the JDK to use unlimited cryptography, set the crypto.policy to a value of 'unlimited'. See the notes in the java.security file shipping with this release for more information.


commitay:
If anyone needs this for an older JDK 8 I think it can be hosted in a personal tap.

Anyone that runs `brew cask install java8 jce-unlimited-strength-policy8` will get an install error.

lhazlewood:
Why was this removed?  This is still needed by so many in the Java community.  Is it possible to reinstate a working version for the `java8` recipe?

commitay:
It doesn't work with the current `java8` Cask.

lhazlewood:
Why not?  What is necessary to make it work with that cask?

commitay:
Please read the top post.

lhazlewood:
Ah, that's right - for 8u152 it's included automatically, but you have to change a property.  Thanks.

Would it be prudent to provide a cask that makes that config file change for the `java8` cask?  Or should the `java8` cask be updated to do that automatically?  Thoughts?

commitay:
> Would it be prudent to provide a cask that makes that config file change for the `java8` cask?  Or should the `java8` cask be updated to do that automatically?  Thoughts?

This is a configuration option that should be set by the user.

lhazlewood:
Why should they be forced to?

I'm asking because the vast majority of people that used the used the previous cask wanted to enable the feature as part of an automated installation.  People don't *want* to modify config properties of the JDK installation.  They want to run a brew command and move on about their lives.  Is this not something we can do for them?

commitay:
You're assuming that everyone that installs the `java8` would want this enabled.

lhazlewood:
I understand what you're saying now.  In that case, what if we reinstate the `jce-unlimited-strength-policy8` cask that makes the file modification for you?  Then it's an opt-in thing.  Thoughts?

commitay:
No, we don't have and don't support Casks that only make configuration changes. Casks require a URL / download.

lhazlewood:
Thanks for the clarification, much appreciated.

PR #4657, opened 2017-10-24
Rename android-studio-canary to android-studio-preview

commitay:
> https://developer.android.com/studio/preview/index.html
> The Android Studio Preview can run side-by-side with your stable version, so you can work on the same projects in both versions.

Upstream seems to call this "Preview". canary, beta, rc are just different releases.

> https://developer.android.com/studio/preview/install-preview.html
> If you download version 2.3 or lower, the application name does not include the version number, so you must first rename it before moving the new version into your apps directory. Otherwise, you might override your existing version of Android Studio.

Supersedes https://github.com/caskroom/homebrew-versions/pull/4656


mxl:
Awesome!

PR #4668, opened 2017-10-27
Update java6.rb

miccal:
The `url` is un-versioned and there is no `appcast`, so the `version` should be `:latest`.

### DIFF
--- a/Casks/java6.rb
+++ b/Casks/java6.rb
@@ -1,6 +1,6 @@
 cask 'java6' do
-  version '1.6.0_65-b14-468'
-  sha256 'a6ea47965542b5c06787f832f3ae5be65da6c6ed91664b0c0ed4994650d98244'
+  version :latest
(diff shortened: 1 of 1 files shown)


commitay:
Disagree, this hasn't been updated for two years and probably won't be again for a while.

Ping @vitorgalvao

vitorgalvao:
Since most users prefer casks to have a `sha256`, I’m partial to leaving it in in casks that don’t change regularly, even if we don’t have an `appcast`.

I’d go so far as to say that in this repo (versions), *all old versions* should have a `version` and `sha256`.

vitorgalvao:
The question to ask is: if a cask has a `version` and `sha256`, do we gain anything by removing it? In this case, the answer is “no”. This will never be updated, and in the off chance it is, it’ll be a rare event that won’t take much work.

PR #4823, opened 2017-11-20
Update google-chrome-beta.rb

miccal:
Pending https://github.com/caskroom/homebrew-cask/pull/41062


commitay:
@vitorgalvao I'm not sure about versioning the beta and dev Chrome Casks, their update frequency doesn't work well with a unversioned `url`.

vitorgalvao:
I’m all for making betas/nightlies *less* versioned and with `url do` blocks. It’s unreasonable to expect we could be on top of all those manually.

PR #5061, opened 2018-01-07
firefoxdeveloperedition: target latest version without checksum

lucafavatella:
For reference - last formula update is 58.0b7 at end Nov 2017, current is version 58.0b14.

I am not sure whether there are distinct best practices for dealing with this kind of formulas.

----

- [x] `brew cask audit --download firefoxdeveloperedition` is error-free.
- [x] `brew cask style --fix firefoxdeveloperedition` reports no offenses.

### DIFF
--- a/Casks/firefoxdeveloperedition.rb
+++ b/Casks/firefoxdeveloperedition.rb
@@ -1,55 +1,44 @@
 cask 'firefoxdeveloperedition' do
-  version '58.0b7'
+  version :latest
+  sha256 :no_check
 
   language 'cs' do
-    sha256 'adf6c02447e5961d9a689ff7190d70bebe0b424b3ee49916606f078ca6e2256f'
     'cs'
   end
 
   language 'de' do
-    sha256 '2d28977006bc504fa47dd71030e4890d31e7559f6d4351c47bde95bb131c46ef'
     'de'
   end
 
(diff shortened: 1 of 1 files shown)


commitay:
We may want to make the same change to `firefox-beta`, weekly updates for Casks with `language` stanzas is hard to keep up with.

Casks/firefoxdeveloperedition.rb
```diff
@@ -1,55 +1,45 @@
 cask 'firefoxdeveloperedition' do
-  version '58.0b7'
+  # Prefer latest version, even if versioned downloads are offered, because of frequent releases.
```
commitay: Remove the comment.

lucafavatella: oh... ok. I was concerned that the formula would bounce back and forth from versioned to unversioned (as happened in the past) though I will remove comment - ok thanks.

lucafavatella:
In force pushed commit, I moved context of change from PR description to commit message, so future formula updaters can see what the rationale behind this change was.

I am running more local checks e.g. `brew cask audit --download firefoxdeveloperedition`.

lucafavatella:
> I am running more local checks e.g. brew cask audit --download firefoxdeveloperedition.

Done.

PR #5240, opened 2018-02-12
Remove gpg-suite-beta

commitay:
This has been tracking the stable release for the last couple of versions as there isn't a beta.


commitay:
Download of `nightly` times out.

PR #5385, opened 2018-03-16
Use explicit URI#open: 10 Casks

commitay:
https://github.com/caskroom/homebrew-versions/pull/5375


commitay:
No, I'm an idiot. I opened the `caskroom/cask` PR twice and didn't open the `versions` PR. Thought it was only a docs update. 😆

reitermarkus:
> No, I'm an idiot.

Well, and here I thought I was tired. Glad we cleared that up. 😛

PR #5411, opened 2018-03-21
Delete java6.rb

vitorgalvao:
No cask in the main repo requires java 6 (they’re at the very least 6+). Let’s remove it and see if there are complaints.


miccal:
Please re-add this, this is the one provided by Apple and a lot of Apps on MacOS only work correctly with this version of Java.

PR #5480, opened 2018-04-05
Update 2 java Casks: comment out `auto_updates`

commitay:


reitermarkus:
I assume this is to prevent further PRs? It should be `auto_updates false`, to be perfectly clear. 😉

commitay:
Yes, we've removed it previously and gotten several PRs trying to add it back.

We don't use `auto_updates false` so I thought commenting the stanza we would use is clearer.

PR #5527, opened 2018-04-17
Update openbazaar-beta to 2.1.1-rc4

gobinathm:


gobinathm:
Isn't Versions for hosting alternative version to the current release. If this cask is going to have 2.1.1 then we should remove this cask altogether. Because this is no longer different from https://github.com/caskroom/homebrew-cask/blob/master/Casks/openbazaar.rb. Basically we have 2 cask's installing same versions.

commitay:
https://github.com/caskroom/homebrew-cask/issues/31772

PR #5530, opened 2018-04-17
Remove java9

commitay:
Download 404s and it has been removed from http://www.oracle.com/technetwork/java/javase/overview/index.html

~~Homepage still works so leaving this on hold for now.~~
https://www.oracle.com/technetwork/java/javase/downloads/jdk9-downloads-3848520.html

Homepage now has:
> Java SE 9 has reached end of support. Users of Java SE 9 should switch to Java SE 10.


joakimortbrant:
Java 9 is in archive
http://www.oracle.com/technetwork/java/javase/downloads/java-archive-javase9-3934878.html

PR #5575, opened 2018-04-22
Update homepage for omnipresence-beta

gtmgianni:
This is the equivalent for what has been set as the homepage for [`omnifocus-beta`](https:///homebrew-versions/Casks/omnifocus-beta.rb).


vitorgalvao:
The download URL itself if 404ing.

gtmgianni:
@vitorgalvao There isn't currently a beta release for OmniPresence – the URL has been 404ing for a few weeks now. I just thought it would be a good idea to update the homepage URL to match the format seen in `omnifocus-beta`, to make it easier in the future to check if there are any updates to the application.

PR #5686, opened 2018-05-15
 Update 2 Casks: change to license caveat DSL

commitay:
Need to wait for a new tag before we merge this.

### DIFF
--- a/Casks/java8.rb
+++ b/Casks/java8.rb
@@ -70,18 +70,16 @@ cask 'java8' do
              ],
       rmdir: '~/Library/Application Support/Oracle/'
 
-  caveats <<~EOS
-    This Cask makes minor modifications to the JRE to prevent issues with
-    packaged applications, as discussed here:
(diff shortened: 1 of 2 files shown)


reitermarkus:
Maybe lose the parentheses, since we don't use parentheses for stanzas either.

commitay:
> Maybe lose the parentheses, since we don't use parentheses for stanzas either.

I don't understand what this means?

commitay:
Okay, I've had another ☕️. If parentheses aren't required (and they aren't) all of our caveats are wrong.

vitorgalvao:
> If parentheses aren't required (and they aren't) all of our caveats are wrong.

Not wrong, but we’re somewhat inconsistent. We always use parenthesis in `depends_on_java`, but never on the other cases (except [one that slipped through](https://github.com/caskroom/homebrew-cask/pull/47286)).

PR #5779, opened 2018-06-04
Delete intellij-idea20161.rb

leipert:
We discussed it some time ago, we brought it back:
https://github.com/Homebrew/homebrew-cask-versions/pull/3714

Is now the right time to get rid of it?

> This is well covered in the acceptable casks policy. For at least one year, this cask has a place here.


vitorgalvao:
After we add analytics, we’ll be able to make better informed decision in these cases. For now, let’s stick to the rules.

PR #5822, opened 2018-06-14
Add eXist-db.app v5.0.0-RC1

adamretter:
I am afraid that I do not have a Mac with a decent enough Internet connection to be able to double-check everything for your task list. Hopefully it is okay though. I based this on a previous PR https://github.com/Homebrew/homebrew-cask-versions/pull/3066 from @joewiz

PR #5828, opened 2018-06-15
contributing: add examples of tokens

commitay:
Mentioned in https://github.com/Homebrew/homebrew-cask-versions/pull/5822#issuecomment-397522013


commitay:
We probably want to keep the PR template pointing to the token reference so contributors are linked to all the naming conventions. I've opened a PR adding links to the token reference for tap specific naming.

https://github.com/Homebrew/homebrew-cask/pull/48553

PR #5984, opened 2018-07-20
Add java-11-ea version 11 early access

jgoldhammer:

- [ x] `brew cask audit --download {{cask_file}}` is error-free.
- [ x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [ x] The commit message includes the cask’s name and version.

Additionally, if **adding a new cask**:

- [ x] Named the cask according to the [token reference].
- [ x] `brew cask install {{cask_file}}` worked successfully.
- [ x] `brew cask uninstall {{cask_file}}` worked successfully.
- [ x] Checked there are no [open pull requests] for the same cask.
- [ x] Checked the cask was not already refused in [closed issues].
- [ x] Checked the cask is submitted to [the correct repo].

[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/Homebrew/homebrew-cask-versions/pulls
[closed issues]: https://github.com/Homebrew/homebrew-cask-versions/issues?q=is%3Aissue+is%3Aclosed
[the correct repo]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask
[version-checksum]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256

### DIFF
--- a/Casks/java11-ea.rb
+++ b/Casks/java11-ea.rb
@@ -1,0 +1,69 @@
+cask 'java11-ea' do
+  version '11-ea+23'
+  sha256 '7bfbcdcfdf8522faef9e389f13ae3a483a9581a2839dd9f0da5f558b33be2e56'
+
+  url "https://download.java.net/java/early_access/jdk11/23/BCL/jdk-#{version}_osx-x64_bin.dmg"
+  name 'Java Standard Edition Development Kit 11'
+  homepage 'http://jdk.java.net/11/'
+
+  # auto_updates true: JDK does not auto-update
+  pkg 'JDK 11.pkg'
+
+  postflight do
+    system_command '/usr/libexec/PlistBuddy',
+                   args: ['-c', 'Add :JavaVM:JVMCapabilities: string BundledApp', "/Library/Java/JavaVirtualMachines/jdk-#{version.split('-')[0]}.jdk/Contents/Info.plist"],
+                   sudo: true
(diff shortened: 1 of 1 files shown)


jgoldhammer:
Hi, thanks for merging.
Can we please revert to the original cask name java11-ea instead of java-beta?

Background:
When installing and updating macOS applications, you normally want to run the latest version or a certain beta of this software. This is the normal case.

In the java world you want to have several java development kit versions on your machine and you want to control which versions are installed. When you want to have java 11, you do not want to install java-beta because this could change later. You may have following situation:
- You have installed java 10 with the normal brew cask java.
- You want to test java 11 and java 12. With the terminology we can only test java 11 as java-beta or later java 12 as java-beta, but not both. Oracle will develop versions in parallel in future and we may want to test our software with several versions.

What do you think how to scope with this situation here?
Should I provide my own tap repo for this special case?

commitay:
> Can we please revert to the original cask name java11-ea instead of java-beta?

> You want to test java 11 and java 12. With the terminology we can only test java 11 as java-beta or later java 12 as java-beta, but not both. 

11 will be GA in 2 months and there isn't an Oracle JDK 12 build available currently?

PR #6134, opened 2018-08-23
cask-versions -> homebrew-cask-versions

brianmorton:


commitay:
`brew tap homebrew/cask-versions` this is correct.

commitay:
`homebrew-` is prefix automatically.

PR #6531, opened 2018-11-12
Update iterm2-beta to 3.2.6beta3

kenhungkk:

### DIFF
--- a/Casks/iterm2-beta.rb
+++ b/Casks/iterm2-beta.rb
@@ -1,7 +1,7 @@
 cask 'iterm2-beta' do
   # note: "2" is not a version number, but an intrinsic part of the product name
   version '3.2.6beta3'
-  sha256 '0b76f0832685a97ca341c7a665c9b20e614bfeee208f031372ab5fbe16412b3d'
+  sha256 '2993719b4e9493e3291bf6357b9a44d9dd35d6004d24493dfcf0ee3b4d3dc626'
 
   url "https://iterm2.com/downloads/beta/iTerm2-#{version.dots_to_underscores}.zip"
   appcast 'https://iterm2.com/appcasts/testing3.xml'


dmofot:
SHA confirmation listed in PR #6533

vitorgalvao:
@dmofot Please don’t tell people to rely on the upstream page. That is *not* sufficient verification for us. If an attacker could have changed the download, they could have faked that information as well.

micdah:
@vitorgalvao How about reverting the available cask to the last know good version then? Seems a bit silly to let it linger here in a state where the available version will fail, meaning anyone wanting to install the beta version through `brew` will be unable to do so.

If verification cannot be obtained from the developer, we'll just have to wait for the next version where an update will fix this release.

micdah:
I have tweeted to the developer of iTerm, let's see if he can confirm the new SHA for us. https://twitter.com/micdah/status/1062583569065930752

vitorgalvao:
> How about reverting the available cask to the last know good version then?

That’d be hell to manage if we did that every time.

dmofot:
@gnachman would it be possible for you to publicly confirm the `sha256` of the iTerm2 3.2.6beta3?

> Ask the developer to publicly confirm the new checksum and provide a link to this confirmation in your pull request. As examples, confirmation can be acheived by opening an issue of the project’s GitHub, or asking on the developer’s Twitter. Note that a link to an upstream page with the checksum is never sufficient, as a malicious third-party who could modify the download could also modify checksums on the page.

vraravam:
> > How about reverting the available cask to the last know good version then?
> 
> That’d be hell to manage if we did that every time.

@vitorgalvao - when comparing that to the "broken" experience for all developers, would that be more or less acceptable? imo, end-user experience trumps the reason that you cited in this case.

vitorgalvao:
> when comparing that to the "broken" experience for all developers

There’s no “broken experience for all developers”. HBC’s update model assumes and encourages you to accept the app’s auto-update mechanism, which still works fine. iTerm’s is so aggressive, even if you download the stable version they ask you to upgrade to the beta (realistically, that’s a bug).

> end-user experience trumps the reason that you cited in this case.

What I ask you to understand is that as non-maintainers there’s a tendency to keep suggesting solutions for *your* problem that doesn’t take the whole into account. As I said, [this would be hell to manage](https://github.com/Homebrew/homebrew-cask-versions/pull/6531#issuecomment-438776192) if we did it every time for every cask. If we as maintainers can’t manage things, you all lose.

We have a system. It’s not perfect and we’re working to improve it, but it’s the best we can work with right now. The time to suggest something different has already passed. The conversation was open, as all our issues are and we picked *something*. Coming in now to to suggest something different because it’s specifically convenient for you while ignoring the whole HBC system is unworkable. *That* would lead to an even worse user experience.

vraravam:
Rather than continuing debating about this, I found a workaround (might be useful for those who face the same issue until the working version is committed/merged). Use the following command to install the previous working version of `iTerm2-beta`:
```bash
brew cask install https://raw.githubusercontent.com/Homebrew/homebrew-cask-versions/ee1caf07db2a2d488f11f406807123e10210937f/Casks/iterm2-beta.rb
```

PR #6949, opened 2019-02-06
Remove workarounds in Java casks.

reitermarkus:
Depends on https://github.com/Homebrew/brew/pull/5686.

All manually verified locally.

### DIFF
--- a/Casks/java8.rb
+++ b/Casks/java8.rb
@@ -6,36 +6,14 @@ cask 'java8' do
       cookies: {
                  'oraclelicense' => 'accept-securebackup-cookie',
                }
-  name 'Java 8 Standard Edition Development Kit'
+  name 'Oracle Java 8 Standard Edition Development Kit'
   homepage 'https://www.oracle.com/technetwork/java/javase/overview/index.html'
 
-  # auto_updates true: JDK does not auto-update
   depends_on macos: '>= :yosemite'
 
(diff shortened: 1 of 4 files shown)


Casks/java8.rb
```diff
@@ -6,36 +6,14 @@
...
   homepage 'https://www.oracle.com/technetwork/java/javase/overview/index.html'
 
-  # auto_updates true: JDK does not auto-update
   depends_on macos: '>= :yosemite'
 
   pkg 'JDK 8 Update 202.pkg'
 
-  postflight do
-    system_command '/bin/ln',
-                   args: ['-nsf', '--', "/Library/Java/JavaVirtualMachines/jdk-#{version.before_comma}.jdk/Contents/Home", '/Library/Java/Home'],
```
Rob--W: What is the reason for removing those symlinks? The removal of the `/Library/Java/Home` symlink broke my workflow (because the unversioned path was hardcoded in build configuration that I used).

reitermarkus: Firstly, to remove the maintenance burden of keeping all Java casks in sync. Secondly, this symlink conflicts between every installed Java version, so there is no guarantee to which version it is pointing anyways. Thirdly, using `/usr/bin/java` and `/usr/libexec/java_home` should be preferred to find the correct Java version (so as to not break in the future) .

PR #6965, opened 2019-02-10
Update iterm2-nightly to 3_3_20190210

gnachman:

### DIFF
--- a/Casks/iterm2-nightly.rb
+++ b/Casks/iterm2-nightly.rb
@@ -1,7 +1,7 @@
 cask 'iterm2-nightly' do
   # note: "2" is not a version number, but an intrinsic part of the product name
-  version :latest
-  sha256 :no_check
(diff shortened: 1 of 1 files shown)


reitermarkus:
This doesn't have an appcast nor a versioned URL.

vitorgalvao:
We could use `https://github.com/gnachman/iTerm2/releases` as an `appcast`.

However, even though you’re the developer, @gnachman, we prefer to have nightlies unversioned, as otherwise the constant barrage of updates would make updating them too problematic.

gnachman:
I think my script is broken anyway, so I'll remove the cask-repair command from it.

vitorgalvao:
@gnachman But if you have a script that submits the stable iTerm2 versions to the main repo (https://github.com/Homebrew/homebrew-cask) as they come out, those PRs we’ll be happy to receive!

PR #7314, opened 2019-05-06
Add Kitematic version 0.17.6

kelbyers:
Install Kitematic version 0.17.6.  Kitematic version 0.17.7 has an
unfixed bug that prevents it from displaying any logs from a containter.
The recommended workaround is to uninstall that version and to install
an older version. Kitematic version 0.17.6 does not have that particular
issue.

- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Additionally, if **adding a new cask**:

- [x] Checked the cask was not [already refused].
- [x] Checked the cask is submitted to [the correct repo].

[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/Homebrew/homebrew-cask-versions/pulls
[already refused]: https://github.com/Homebrew/homebrew-cask-versions/search?q=is%3Aclosed&type=Issues
[the correct repo]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask
[version-checksum]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256

### DIFF
--- a/Casks/kitematic0176.rb
+++ b/Casks/kitematic0176.rb
@@ -1,0 +1,22 @@
+cask 'kitematic0176' do
+  version '0.17.6'
+  sha256 '827756c4b7de7b23b474aba12ed2b4a333d3a5f1b2fad5bc5819da405d72ed83'
+
+  # github.com/docker/kitematic was verified as official when first introduced to the cask
+  url "https://github.com/docker/kitematic/releases/download/v#{version}/Kitematic-#{version}-Mac.zip"
+  appcast 'https://github.com/docker/kitematic/releases.atom'
+  name 'Kitematic'
+  homepage 'https://kitematic.com/'
+
+  app 'Kitematic.app'
+
+  zap trash: [
+               '~/Kitematic',
+               '~/Library/Application Support/Kitematic',
+               '~/Library/Caches/Kitematic',
+               '~/Library/Logs/Kitematic',
+               '~/Library/Preferences/com.electron.kitematic.plist',
+               '~/Library/Preferences/com.electron.kitematic.helper.plist',
+               '~/Library/Saved Application State/com.electron.kitematic.savedState',
+             ]
+end


vitorgalvao:
I missed this; apologies.

However, this doesn’t seem to be an issue affecting anyone else, and since this repo [has stricter inclusion rules](https://github.com/Homebrew/homebrew-cask-versions#acceptable-casks), I’ll refuse this one.

Thank you anyway.

kelbyers:
I am not sure that I understand why this was rejected. The current version of this application literally does not work, and the only workaround that has been found is to install a version of 0.17.6 or older. The issue is not even limited to Mac, but also effects Ubuntu and Windows, based on comments in https://github.com/docker/kitematic/issues/4610.

Since `brew cask` does not support installing arbitrary versions of an application, the two alternatives for getting a useful version are to not use homebrew at all for this application, or to have a version that works be available in brew-cask-versions.

My request for including this is not because I specifically like this version best, but because this version is the most recent one that actually works.

vitorgalvao:
> Since `brew cask` does not support installing arbitrary versions of an application, the two alternatives for getting a useful version are to not use homebrew at all for this application, or to have a version that works be available in brew-cask-versions.

Or you can host your own [tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md), or install a specific git revision of the cask. The latter is not something we support officially, but it does work, and this regression should be temporary (meaning soon the cask will be lingering here with no purpose). For such a popular cask (in the main repo), it would be expected that more people would be asking for this/submitting it. That is usually the case but it’s not happening here, which leads me to believe this is not a huge issue amongst our users.

But I’ll let it in and let the regular purge do its thing when the time comes. There will be plenty of time.

PR #7370, opened 2019-05-19
Secure URLs

vszakats:

- [ ] The commit message includes the cask’s name and version.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

---
```
2 files inspected, no offenses detected
```
```
==> Downloading https://github.com/aonez/Keka/releases/download/v1.1.5-rc.2/Keka-1.1.5-rc.2.dmg
==> Downloading from https://github-production-release-asset-2e65be.s3.amazonaws.com/73220421/7dad5c00-d9c8-11e8-8e2c-346428811823?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Am
==> Verifying SHA-256 checksum for Cask 'keka-beta'.
audit for keka-beta: passed
==> Downloading https://downloads.sourceforge.net/deadbeef/travis/osx/master/deadbeef-devel-osx-x86_64.zip
==> Downloading from https://liquidtelecom.dl.sourceforge.net/project/deadbeef/travis/osx/master/deadbeef-devel-osx-x86_64.zip
==> No SHA-256 checksum defined for Cask 'deadbeef-devel', skipping verification.
audit for deadbeef-devel: passed
```

### DIFF
--- a/Casks/deadbeef-devel.rb
+++ b/Casks/deadbeef-devel.rb
@@ -2,9 +2,10 @@ cask 'deadbeef-devel' do
   version :latest
   sha256 :no_check
 
+  # downloads.sourceforge.net/deadbeef was verified as official when first introduced to the cask
   url 'https://downloads.sourceforge.net/deadbeef/travis/osx/master/deadbeef-devel-osx-x86_64.zip'
   name 'DeaDBeeF'
-  homepage 'http://deadbeef.sourceforge.net/'
+  homepage 'https://deadbeef.sourceforge.io/'
 
   app 'deadbeef.app'
 
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -68,7 +68,7 @@ members of the project's leadership.
 ## Attribution
 
 This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
-available at [http://contributor-covenant.org/version/1/4][version]
+available at [https://www.contributor-covenant.org/version/1/4][version]
 
-[homepage]: http://contributor-covenant.org
-[version]: http://contributor-covenant.org/version/1/4/
+[homepage]: https://www.contributor-covenant.org/
+[version]: https://www.contributor-covenant.org/version/1/4/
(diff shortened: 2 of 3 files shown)


Casks/deadbeef-devel.rb
```diff
@@ -2,9 +2,10 @@
   version :latest
   sha256 :no_check
 
+  # downloads.sourceforge.net/deadbeef was verified as official when first introduced to the cask
   url 'https://downloads.sourceforge.net/deadbeef/travis/osx/master/deadbeef-devel-osx-x86_64.zip'
   name 'DeaDBeeF'
-  homepage 'http://deadbeef.sourceforge.net/'
+  homepage 'https://deadbeef.sourceforge.io/'
```
vitorgalvao: This redirects to the old non-HTTPS URL. That being the case, I’d rather keep it as it was.

vszakats: Last few times we've discussed this, Cask maintainers agreed that a redirect to insecure for a homepage isn't of any concern. We do use the `.io` flavour for every Homebrew core formulae as well. Enabling security is just a switch on the sf.net control panel that the project admin can do (this may also need fixing HTML mixed-content issues.)

vitorgalvao: That doesn’t apply when the `was verified as official when first introduced to the cask` comment is required, which is why I only commented on those cases.

vszakats: This doesn't make sense to me.

vitorgalvao: > This doesn't make sense to me.

When we add that comment, we’re increasing overhead. It increases the burden of auditing casks because it needs to be done manually. To do it *for no gain at all* (since the secure home address redirects to an insecure one) makes little sense.

PR #7448, opened 2019-06-05
Update safari-technology-preview.rb from 83 to 84

suschizu:
this will surely cause an error


suschizu:
`Cask safari-technology-preview depends on macOS release >= 10.14, but you are running release 10.13.`
Mojave and Catalina …

PR #7477, opened 2019-06-13
Update safari-technology-preview.rb

core-code:
for some reason i can't send it through cask-repair
```

• == /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask-versions/Casks/safari-technology-preview.rb ==
C:  6:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.
C:  9:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.

1 file inspected, 2 offenses detected
• == /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask-versions/Casks/safari-technology-preview.rb ==
C:  6:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.
C:  9:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.

1 file inspected, 2 offenses detected
• == /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask-versions/Casks/safari-technology-preview.rb ==
C:  6:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.
C:  9:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.

1 file inspected, 2 offenses detected
• == /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask-versions/Casks/safari-technology-preview.rb ==
C:  6:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.
C:  9:  5: Move sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a' out of the conditional.

```

### DIFF
--- a/Casks/safari-technology-preview.rb
+++ b/Casks/safari-technology-preview.rb
@@ -1,9 +1,15 @@
 cask 'safari-technology-preview' do
-  version '85,041-73105-20190612-8d1ff016-e901-431d-aa35-7cf7b1ceaa7d'
-  sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a'
+  version '85,041-73105-20190612-8d1ff016-e901-431d-aa35-7cf7b1ceaa7d/'
 
-  url "https://secure-appldnld.apple.com/STP/#{version.after_comma}/SafariTechnologyPreview.dmg"
-  appcast 'https://developer.apple.com/safari/technology-preview/release-notes/'
+  if MacOS.version <= :mojave
+    url 'https://secure-appldnld.apple.com/STP/041-73107-20190612-2744e373-b7d6-47b2-984c-08bcf11661ab/SafariTechnologyPreview.dmg'
+    sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a'
+  else
+    url "https://secure-appldnld.apple.com/STP/#{version.after_comma}/SafariTechnologyPreview.dmg"
+    sha256 '36c743f5d5442772ca137922818f98cc07e5d72e63ff5181df6f4fe0d6cd5ef5'
+  end
+
+  appcast 'https://developer.apple.com/safari/download/'
   name 'Safari Technology Preview'
   homepage 'https://developer.apple.com/safari/download/'
 
--- a/Casks/safari-technology-preview.rb
+++ b/Casks/safari-technology-preview.rb
@@ -1,5 +1,5 @@
 cask 'safari-technology-preview' do
-  version '85,041-73105-20190612-8d1ff016-e901-431d-aa35-7cf7b1ceaa7d/'
+  version '85,041-73105-20190612-8d1ff016-e901-431d-aa35-7cf7b1ceaa7d'
 
   if MacOS.version <= :mojave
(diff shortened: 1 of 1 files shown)


vitorgalvao:
It’s a `style` error. The shasum is repeated.

core-code:
huh? the SHA needs to be twice and always has been in this cask. am i missing something here?

core-code:
i don't think removing the conditional is a good idea...?

vitorgalvao:
Both branches of the conditional were pointing at the same file, thus getting the same checksum. This is because the first conditional of the cask was now `<= :mojave`, which is everything (stable); the `else` would only trigger in Catalina.

core-code:
thanks for pointing that out. i think it should be good now

core-code:
i think the only remaining failure is expected, travis still runs on 10.13

Casks/safari-technology-preview.rb
```diff
@@ -1,15 +1,15 @@
...
+    url 'https://secure-appldnld.apple.com/STP/041-73107-20190612-2744e373-b7d6-47b2-984c-08bcf11661ab/SafariTechnologyPreview.dmg'
+    sha256 '52c70d1f62c4d68d93e371c034fa37395bdd63dd09fcc8037679e01ccd46b68a'
   else
     url "https://secure-appldnld.apple.com/STP/#{version.after_comma}/SafariTechnologyPreview.dmg"
-    sha256 '539f6df5a18122e91ff515446858ad310ba42b9057dc9385ee36e6ade2275587'
+    sha256 '36c743f5d5442772ca137922818f98cc07e5d72e63ff5181df6f4fe0d6cd5ef5'
   end
 
-  appcast 'https://developer.apple.com/safari/technology-preview/release-notes/'
+  appcast 'https://developer.apple.com/safari/download/'
```
foolip: Was this change intentional? https://developer.apple.com/safari/technology-preview/release-notes/ still exists, and this URL is already used for `homepage`.

https://github.com/Homebrew/homebrew-cask-versions/pull/7485 was autocreated because of the difference.

core-code: > Was this change intentional? 

yes

> https://developer.apple.com/safari/technology-preview/release-notes/ still exists

yes it exists, but that is no reason to have it in the appcast stanza

> and this URL is already used for homepage.

yes, but we have to use a URL in the appcast stanza that makes sense, whether its already in the homepage URL too doesn't matter

the reason for this change is that we use the appcast stanzas in various semi-automated ways to keep casks up-to-date. this is not possible with the release notes url that was in there previously.

i don't mind if you change it back, if you promise to keep this cask up to date in a timely manner ;)

foolip:
I'll adapt my updater to match this, just wanted to make sure it wasn't a mistake.

My updater isn't perfectly reliable, so until it is I don't want to promise anything timely :)

core-code:
ok sounds good. sorry for causing disturbances on your end there ...

foolip:
Alright, updated to match.

PR #7599, opened 2019-07-10
Update safari-technology-preview.rb from 86 to 87

suschizu:
❤️ ❤️ ❤️


suschizu:
`Cask safari-technology-preview depends on macOS release >= 10.14, but you are running release 10.13.`
thats because of catalina …

PR #7655, opened 2019-07-24
Update dotnet-sdk-preview to latest preview

YaourtGG:
From version 3.0.100-preview6-012264 to version 3.0.100-preview7-012821

### DIFF
--- a/Casks/dotnet-sdk-preview.rb
+++ b/Casks/dotnet-sdk-preview.rb
@@ -1,8 +1,8 @@
 cask 'dotnet-sdk-preview' do
-  version '3.0.100-preview7-012821'
+  version '3.0.100-preview7-012821,64cb8405-ee15-4a9a-bf25-1201531f4519/b619596c137a08b204fc79a213bb9763'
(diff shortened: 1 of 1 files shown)


core-code:
i guess we should append '64cb8405-ee15-4a9a-bf25-1201531f4519/b619596c137a08b204fc79a213bb9763' to the version

YaourtGG:
Why would we ?
I think it has nothing to do with dotnet's current version.
Furthermore, this would result in a really long version name.

vitorgalvao:
> Why would we ?

So it can be interpolated in `url`.

> I think it has nothing to do with dotnet's current version.

We care for maintainability more than the exact version. Almost no one cares about what it says on the version.

> Furthermore, this would result in a really long version name.

And faster PR merges, and automatability, and less things to change on updated, and fewer errors, and it’s not the first cask to do it. @core-code’s suggestion is our preferred method.

core-code:
> > I think it has nothing to do with dotnet's current version.
> Almost no one cares about what it says on the version.

well, i do, but its not an either-or, we can always carry both the actual app version and the download token as is done now. this makes everyone happy ;)

PR #7722, opened 2019-08-08
Add textmate-unstable

core-code:
As discussed here:

https://github.com/Homebrew/homebrew-cask/pull/67190

### DIFF
--- a/Casks/textmate-latest.rb
+++ b/Casks/textmate-latest.rb
@@ -1,0 +1,28 @@
+cask 'textmate-latest' do
+  version '2.0-rc.29'
+  sha256 '49c14380728d8bf1af1e09298fa616a09ae535f5f106967361b84f6ec81d6249'
+
+  # github.com/textmate/textmate was verified as official when first introduced to the cask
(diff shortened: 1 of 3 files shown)


Casks/textmate-latest.rb
```diff
@@ -0,0 +1,28 @@
+cask 'textmate-latest' do
```
vitorgalvao: `latest` is ambiguous for the mais repo. Would `unstable` make sense?

core-code: you are right, that makes sense. i can rename once i am in front of a computer

reitermarkus: How about `textmate-preview`? The preference `.plist` was called `com.macromates.TextMate.preview` at some point in the past at least.

core-code: i don't have an opinion here. whatever you decide, let me know and i'll rename accordingly.

vitorgalvao: I’m fine with `preview`.

PR #7992, opened 2019-10-14
remove adoptopenjdk8

ryenus:
As suggested by @reitermarkus:
 - https://github.com/Homebrew/homebrew-cask/issues/57387#issuecomment-533295298

> > causing a conflict with cask-versions on adoptopenjdk8

> I think there's an open issue for this bug. If there isn't one, someone should open one, or better yet, submit a PR fixing it.

It's better to remove this cask to avoid conflict with the adoptopenjdk tap:
 - https://github.com/AdoptOpenJDK/homebrew-openjdk/


reitermarkus:
> As suggested by @reitermarkus 

I was suggesting fixing the core code so it isn't treated as a conflict in the first place. This only hides the problem but doesn't fix the root cause.

ryenus:
@reitermarkus, thanks for clarifying, I'll close this then.

PR #8056, opened 2019-10-27
Update java6 to install also on Catalina

nicerloop:

### DIFF
--- a/Casks/java6.rb
+++ b/Casks/java6.rb
@@ -18,11 +18,10 @@ cask 'java6' do
   else
     container nested: 'JavaForOSX.pkg'
 
-    artifact 'Library/Java/JavaVirtualMachines/1.6.0.jdk', target: '/Library/Java/JavaVirtualMachines/1.6.0.jdk'
+    artifact 'JavaForOSX/JavaForOSX.pkg/Payload/Library/Java/JavaVirtualMachines/1.6.0.jdk', target: '/Library/Java/JavaVirtualMachines/1.6.0.jdk'
 
     preflight do
-      system_command 'pkgutil', chdir: staged_path, args: ['--expand', 'JavaForOSX.pkg', 'JavaForOSX']
-      system_command 'tar', chdir: staged_path, args: ['-xzf', 'JavaForOSX/JavaForOSX.pkg/Payload']
+      system_command 'pkgutil', chdir: staged_path, args: ['--expand-full', 'JavaForOSX.pkg', 'JavaForOSX']
     end
   end
 end
--- a/Casks/java6.rb
+++ b/Casks/java6.rb
@@ -7,12 +7,22 @@ cask 'java6' do
   name 'Apple Java 6 Standard Edition Development Kit'
   homepage 'https://support.apple.com/kb/DL1572'
 
-  container nested: 'JavaForOSX.pkg'
+  if MacOS.version <= :mojave
+    pkg 'JavaForOSX.pkg'
 
-  artifact 'Library/Java/JavaVirtualMachines/1.6.0.jdk', target: '/Library/Java/JavaVirtualMachines/1.6.0.jdk'
+    uninstall pkgutil: [
+                         'com.apple.pkg.JavaForMacOSX107',
+                         'com.apple.pkg.JavaMDNS',
+                         'com.apple.pkg.JavaEssentials',
+                       ]
+  else
+    container nested: 'JavaForOSX.pkg'
 
-  preflight do
-    `cd "#{staged_path}" && pkgutil --expand JavaForOSX.pkg JavaForOSX`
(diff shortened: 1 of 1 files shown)


Casks/java6.rb
```diff
@@ -7,11 +7,12 @@
   name 'Apple Java 6 Standard Edition Development Kit'
   homepage 'https://support.apple.com/kb/DL1572'
 
-  pkg 'JavaForOSX.pkg'
+  container nested: 'JavaForOSX.pkg'
 
-  uninstall pkgutil: [
-                       'com.apple.pkg.JavaForMacOSX107',
-                       'com.apple.pkg.JavaMDNS',
-                       'com.apple.pkg.JavaEssentials',
```
reitermarkus: Does this also handle the last two?

Casks/java6.rb
```diff
@@ -7,11 +7,12 @@
...
-                     ]
+  artifact 'Library/Java/JavaVirtualMachines/1.6.0.jdk', target: '/Library/Java/JavaVirtualMachines/1.6.0.jdk'
+
+  preflight do
+    `cd "#{staged_path}" && pkgutil --expand JavaForOSX.pkg JavaForOSX`
```
reitermarkus: Use `system_command`.

Casks/java6.rb
nicerloop: These packages do not appear to be registered on my Mojave machine:

$ brew cask uninstall java6
==> Uninstalling Cask java6
==> Uninstalling packages:
com.apple.pkg.JavaForMacOSX107
Unknown error Error Domain=NSCocoaErrorDomain Code=4 "“com.apple.pkg.JavaForMacOSX107.bom” couldn’t be removed." UserInfo={NSFilePath=/var/db/receipts/com.apple.pkg.JavaForMacOSX107.bom, NSUserStringVariant=(
    Remove
), NSUnderlyingError=0x7fd89f415fa0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}.
==> Purging files for version 1.6.0_65-b14-468 of Cask java6

I have manually searched the pkg and only found a reference to com.apple.pkg.JavaForMacOSX107 which only installs the 1.6.0.jdk artifact.

Casks/java6.rb
nicerloop: Sorry, but I have no experience with ruby other than some copy/paste for some casks maintenance. Could you give me any pointer for the desired modification?

reitermarkus: Look at some other casks using `system_command`, e.g. https://github.com/Homebrew/homebrew-cask/blob/1c980c8b02a3e2deb9d593edf29f15c732c43e49/Casks/pokemon-trading-card-game-online.rb#L14-L15.

You will also need to add the `chdir:` parameter to change the directory.

nicerloop: I found some examples through the repository, made changes that seem to work

Casks/java6.rb
```diff
@@ -18,11 +18,10 @@
   else
     container nested: 'JavaForOSX.pkg'
```
reitermarkus: Why is this necessary?

nicerloop: It extracts the pkg from the dmg

nicerloop: In fact is is not necessary.

Casks/java6.rb
reitermarkus: Since these don't seem to be used anymore, remove them.

Casks/java6.rb
```diff
@@ -7,11 +7,15 @@
...
-                       'com.apple.pkg.JavaForMacOSX107',
-                       'com.apple.pkg.JavaMDNS',
-                       'com.apple.pkg.JavaEssentials',
-                     ]
+    uninstall pkgutil: 'com.apple.pkg.JavaForMacOSX107'
+  else
+    artifact 'JavaForOSX/JavaForOSX.pkg/Payload/Library/Java/JavaVirtualMachines/1.6.0.jdk', target: '/Library/Java/JavaVirtualMachines/1.6.0.jdk'
+
+    preflight do
+      system_command 'pkgutil', chdir: staged_path, args: ['--expand-full', 'JavaForOSX.pkg', 'JavaForOSX']
```
vitorgalvao: Why is this necessary?

nicerloop: The pkg would not install on my Catalina machine, and I'm not the only one to face this case. In fact the pkg just install the jdk bundle in the right location: /Library/Java/JavaVirtualMachines/. Options are 1) patch the pkg Distribution script or 2) expand the pkg and move the jdk bundle to the right location. As both need expanding the pkg, I just went with the second one. Just to be safe, I set this up only for Catalina, but should be applicable to any platform.

PR #8341, opened 2019-12-18
Actions: update to macOS-latest

mistydemeo:
The macOS 10.14 VM has been deprecated, so this needs to update to target the latest version. 10.15 is now the [only supported macOS version](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners).

(I've removed the PR template since this doesn't touch an actual cask. 😄)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   ci:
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Thanks @mistydemeo!

PR #8437, opened 2020-01-20
Update julia-nightly from 1.4 to 1.5

rofinn:
This change includes:

1. Updating julia-nightly from 1.4 to 1.5
2. Making the binary target `julia-nightly` rather than `julia` (to avoid conflicts with the primary julia cask).

I've placed these changes as separate commits in one PR, but I'm happy to squash or make separate PRs if requested.

  - NOTE: The middle commit doesn't change the version, but I can always squash it if you prefer?
- [X] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

### DIFF
--- a/Casks/julia-nightly.rb
+++ b/Casks/julia-nightly.rb
@@ -1,5 +1,5 @@
 cask 'julia-nightly' do
-  version '1.4'
+  version '1.5'
   sha256 :no_check # required as upstream package is updated in-place
 
   url 'https://julialangnightlies-s3.julialang.org/bin/mac/x64/julia-latest-mac64.dmg'
--- a/Casks/julia-nightly.rb
+++ b/Casks/julia-nightly.rb
@@ -7,7 +7,7 @@ cask 'julia-nightly' do
   homepage 'https://julialang.org/'
 
   app "Julia-#{version}.app"
-  binary "#{appdir}/Julia-#{version}.app/Contents/Resources/julia/bin/julia"
+  binary "#{appdir}/Julia-#{version}.app/Contents/Resources/julia/bin/julia", target: 'julia-nightly'
 
   zap trash: '~/.julia'
 end


Casks/julia-nightly.rb
```diff
@@ -1,13 +1,13 @@
...
+  version '1.5'
   sha256 :no_check # required as upstream package is updated in-place
 
   url 'https://julialangnightlies-s3.julialang.org/bin/mac/x64/julia-latest-mac64.dmg'
   name 'Julia'
   homepage 'https://julialang.org/'
 
   app "Julia-#{version}.app"
-  binary "#{appdir}/Julia-#{version}.app/Contents/Resources/julia/bin/julia"
+  binary "#{appdir}/Julia-#{version}.app/Contents/Resources/julia/bin/julia", target: 'julia-nightly'
```
vitorgalvao: Don’t rename the `binary`, add a `conflicts_with` to both casks instead.

rofinn: I'm a little confused by the documentation. Does `conflicts_with` prevent you from using both casks then?

vitorgalvao: > Does `conflicts_with` prevent you from using both casks then?

Prevents you from using both at the same time.

rofinn: Yeah, that's my concern. My ideal use case is being able to run `julia`, `julia-nightly` and `julia-lts` concurrently. Perhaps `brew` isn't the right tool for this though?

vitorgalvao: We’d rather not rename tools provided by upstream. Though admittedly we’re more lenient in what concerns `binary`, and since they change the name of the `app`, I’ll consider it OK for now.

rofinn: Okay, thank you.

rofinn:
If you're okay with this MR would you mind also taking a look at https://github.com/Homebrew/homebrew-cask-versions/pull/8438 when you get a chance? It has the same issue with renaming the `binary` though.

vitorgalvao:
> would you mind also taking a look at #8438 when you get a chance?

I was already aware of both PRs before you pinged me.

> It has the same issue with renaming the `binary` though.

That’s not an issue. Accepting this one I’d also accept the other.

I’m only in doubt if it’s worth including the cask, because it doesn’t seem like it would be nearly popular enough to justify addition. I may accept it for now and revisit in a year’s time.

rofinn:
> I may accept it for now and revisit in a year’s time.

Yeah, that's fair. I imagine the LTS version will only be used by businesses and governments who can't update language versions regularly. I think for most people the regular Julia cask will suffice. If you decide that these casks don't make sense in this tap I'd be happy to setup a separate one.

pthariensflame:
@rofinn If you'd consider adding my `julia-rc` (from #8458) to a separate tap, since I'm not skilled enough to do so myself, I'd be extremely grateful!

vitorgalvao:
> to a separate tap, since I'm not skilled enough to do so myself,

[There’s nothing to it](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

Create a new repo, and name if `homebrew-<something>` (replace `<something>` with whatever you like). Make a `Casks` directory inside and stick your cask file there. Done.

rofinn:
Is there any kind of formal deprecation process we should follow if/when we want to move `julia-lts` and `julia-nightly` to a different tap?

vitorgalvao:
> Is there any kind of formal deprecation process we should follow if/when we want to move `julia-lts` and `julia-nightly` to a different tap?

[Yes](https://docs.brew.sh/Migrating-A-Formula-To-A-Tap), but we don’t use it for taps external to the project. If we remove these casks it would be because they don’t get enough use, anyway.

PR #8481, opened 2020-01-29
Update airmail-beta to 4.0,456

kPherox:
Include `mctoken` in `version` as it is needed when downloading old version.

---

### DIFF
--- a/Casks/airmail-beta.rb
+++ b/Casks/airmail-beta.rb
@@ -1,9 +1,9 @@
 cask 'airmail-beta' do
-  version '4.0,456:1d5be16b070a28f1d94639690cb9764faf862fce'
+  version '4.0,456'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> as it is needed when downloading old version.

What do you mean? What old version?

kPherox:
For example, if you have the current latest `4.0,456` and the cask version is old `4.0,450`, you will need a `mctoken` to download `4.0,450` correctly.

vitorgalvao:
If the cask version is old, it should be updated. This is a beta, not on the main repo, so that it fails in those situations is an advantage, because it will force the cask to be kept up to date, *and* it doesn’t make the cask more complex.

Please revert that part of the change.

kPherox:
Removed token parts

PR #8513, opened 2020-02-06
Update corretto8 from 8.242.08.1 to 8.242.08.1-1

thomasvandoren:

This also required an update to the url calculation, since the trailing `-1` is not included in one part of the url. https://github.com/corretto/corretto-8/releases/tag/8.242.08.1-1


vitorgalvao:
`-1` is usually temporary. We’d always have to change it twice, so might as well make it simple.

PR #8864, opened 2020-04-03
affinity-photo-beta: Updated to 1.8.3.178

thirdtapbot:
cc @vitorgalvao

### DIFF
--- a/Casks/affinity-photo-beta.rb
+++ b/Casks/affinity-photo-beta.rb
@@ -1,5 +1,5 @@
 cask 'affinity-photo-beta' do
-  version '1.8.3.180'
+  version '1.8.3.178'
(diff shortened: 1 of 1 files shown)


core-code:
changing just the version number but not the SHA is almost always wrong (except when correcting past mistakes).

it is also wrong in this case. the version number is definitely 1.8.3.180

vitorgalvao:
@core-code The account is a bot, hence why I allow the direct ping with no more information.

It’s been giving bad results for a while and I’ve been putting off turning it off (no longer worth the fix).

It uses direct information from another feed I’ve built (which works well) that informs about the latest features, but the Affinity developers have been inconsistent in their forums (which use the awful IPBoard).

They have Sparkle feeds for the betas, but Affinity Designer’s is outdated. I’ve contacted its project manager on Twitter a few days ago about it, and he said he was going to try to keep it updated.

core-code:
ah ok, thanks for the explanation

PR #8910, opened 2020-04-10
Update android-studio-preview from 4.1.0.4,193.6325121 to 4.1.0.5,193.6362631

core-code:


core-code:
there are TWO eap versions on the homepage "beta" and "canary". i am updating to the "canary" one because thats whats currently in the cask, but we probably want two different ones?

PR #8955, opened 2020-04-18
Revert "Delete microsoft-remote-desktop-beta.rb"

lumaxis:
Adding microsoft-remote-desktop-beta back with a fixed Formula.

### DIFF
--- a/Casks/microsoft-remote-desktop-beta.rb
+++ b/Casks/microsoft-remote-desktop-beta.rb
@@ -1,0 +1,22 @@
+cask 'microsoft-remote-desktop-beta' do
+  version '10.4.0.1771,531'
+  sha256 '5a2f87ef909074a35364de138164273673d463228c9e104553c1e7eb44039606'
+
+  url "https://rink.hockeyapp.net/api/2/apps/5e0c144289a51fca2d3bfa39ce7f2b06/app_versions/#{version.after_comma}?format=zip"
+  appcast 'https://rink.hockeyapp.net/api/2/apps/5e0c144289a51fca2d3bfa39ce7f2b06',
+          configuration: version.major_minor_patch
+  name 'Microsoft Remote Desktop Beta'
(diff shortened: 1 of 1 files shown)


miccal:
@lumaxis this is not going to work -- if you check [this issue](https://github.com/Homebrew/homebrew-cask/issues/80779) the problem here is that HockeyApp links redirect to App Center links and the redirects are unstable.

lumaxis:
Thanks @miccal, I've read the issue thoroughly 🙂 
I had at first falsely remembered that we had built in support for specific versions in the legacy API support for HockeyApp which would have made this PR work just fine.
Now I just realized that the API for simplicity always returns the latest available version.

That said, the redirect is stable in the way that https://rink.hockeyapp.net/api/2/apps/5e0c144289a51fca2d3bfa39ce7f2b06?format=zip always returns to a working download URL, it just doesn't support requesting specific versions. So an unversioned Cask Formula would theoretically still work.

vitorgalvao:
Hockeyapp is dead and links are starting to redirect to appcenter. Even if this works now, it won’t for long. We need a long-term solution.

lumaxis:
@vitorgalvao I'm very aware, I worked on both and am still on the team at Microsoft that owns App Center. Was just trying to experiment with possible temporary solutions while I try to convince the stakeholders to add a permanent one 😊

vitorgalvao:
Unfortunately, in my experience temporary solutions in casks more often than not become permanent hacks that are abandoned and left for us to fix when they break and no one even remembers them anymore.

If you can get the decision makers to make appcenter URLs that aren’t averse to direct CLI downloads, you’d be helping a lot of casks and their users. But we’re not able to rely on a “may happen”, especially on the versions repo, which is less organised.

lumaxis:
That’s perfectly fair and completely understandable. Part of the motivation for this PR was also my own curiosity and trying to understand again how our own system works 😄

I’ll ask around on the team next week and reach out when I learn more!

PR #9059, opened 2020-05-11
Update android-studio-preview from 4.1.0.8,193.6423924 to 4.1.0.9,201.6466190

ouchadam:


ouchadam:
@vitorgalvao any ideas why we're seeing? 
`The URL https://developer.android.com/studio/preview/ is not reachable`
🤔

vitorgalvao:
@ouchadam Likely because whatever method we use to check if the homepage is live isn’t accepted by the server. A few check for user agents and stuff like that. Whatever method we use works with some, fails with others.

PR #9457, opened 2020-08-13
Update processing2 url

miccal:
Updated the `url` to prevent the `audit` warning:
```
Download uses GitHub releases, please add an appcast.
```
Note that this cask is for an [older version](https://processing.org/download/).


miccal:
The remaining `audit` error should be fixed by https://github.com/Homebrew/homebrew-cask-versions/pull/9458.

PR #9458, opened 2020-08-13
pext-nightly: update url do

miccal:
Change the `url do` block to not use the GitHub API, which (hopefully) fixes the `audit` error:
```
exception while auditing pext-nightly: 403 rate limit exceeded
```


miccal:
The remaining `audit` error should be fixed by https://github.com/Homebrew/homebrew-cask-versions/pull/9457.

PR #9562, opened 2020-09-04
Update osu-development from 2020.820.0 to 2020.903.0

i0ntempest:

- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.

Created with [`cask-repair`](https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#updating-a-cask).

### DIFF
--- a/Casks/osu-development.rb
+++ b/Casks/osu-development.rb
@@ -1,6 +1,6 @@
 cask "osu-development" do
-  version "2020.820.0"
-  sha256 "d277523c3d32a24e8eeaf1d216d690a7000255de2504c69327153ad55e04913d"
+  version "2020.903.0"
+  sha256 "de3fe76bd057c2a72a95db930cd603bb7a08bca822adf6b8e1d7117c1e2aaa70"
 
   url "https://github.com/ppy/osu/releases/download/#{version}/osu.app.zip"
   appcast "https://github.com/ppy/osu/releases.atom"
--- a/Casks/osu-development.rb
+++ b/Casks/osu-development.rb
@@ -7,7 +7,6 @@ cask "osu-development" do
   name "osu!lazer"
   homepage "https://github.com/ppy/osu/"
 
-  conflicts_with cask: "osu"
   depends_on macos: ">= :sierra"
 
   app "osu!.app"


miccal:
The `osu` cask in the main repo [was removed](https://github.com/Homebrew/homebrew-cask/pull/72262), so please remove the line:
```
conflicts_with cask: "osu"
```

i0ntempest:
Done

reitermarkus (approved):

claui:
@reitermarkus I agree. The current token no longer makes sense.

i0ntempest:
For macOS this development version is the only version that runs (aside from the wine version that got removed), but for Windows we have another stable release version. So "osu-development" does make sense, but I do agree that this can be moved to the main repo.

claui:
@i0ntempest What the README tells me:

> This project is under heavy development, but is in a stable state. Users are encouraged to try it out and keep it installed alongside the stable osu! client. It will continue to evolve to the point of eventually replacing the existing stable client as an update.

It says it’s essentially a stable version. The message that comes across to me is that it’s as stable a channel as it gets for macOS.
Mac users shouldn’t really have to concern themselves with osu’s Windows release channels, or how stable these are.

I’m not really familiar with osu so can you help clear this up? In what ways do you feel it’d be useful to keep the `-development` part?

i0ntempest:
@claui This version (so called lazer version) is very different from the stable ver (totally different engine, UI, underlying behavior etc.), and it gets an update every a few days. If we remove ‘-development’ users may think this is the current stable version. However the devs do plan on releasing version as stable after most of the developments are done, so before that, I think we should either keep ‘-development’ or just rename it to ‘osu-lazer’.

core-code:
> If we remove ‘-development’ users may think this is the current stable version

agreed - we shouldn't override upstreams decision in this case and possibly confuse users with that

claui:
Good points @i0ntempest and @core-code. Convinced 👍

i0ntempest:
So should I close this and move this cask to the main repo?

vitorgalvao:
> So should I close this and move this cask to the main repo?

I think the consensus is to keep it in this repo.

Also, I would not be opposed to readding the stable cask to the main repo, as long as we can get official links (homepage and url) that are not from a forum.

i0ntempest:
The stable version is technically Windows only, the "macOS ver" is basically a wineskin wrapper so there's no official link.

Issue #9700, opened 2020-10-02
vlc-nightly doesn't get updated due to new repository structure


yannpub:
Hi,

I just tried to install vlc-nightly. It appears the old directory structure is used : https://nightlies.videolan.org/build/macosx-intel/last/
This structure and URL has changed since beginning of April, and therefore vlc-nightly doesn't get updated.
The new URL is https://artifacts.videolan.org/vlc/nightly-macos/ however there is no "last' folder anymore where the latest version would be stored. New folder structure is now the following:

../
--
20201002-0428/ |   | 02-Oct-2020 04:28
20201001-0428/ |   | 01-Oct-2020 04:28
20200930-0431/ |   | 30-Sep-2020 04:31
20200929-0440/ |   | 29-Sep-2020 04:40
20200928-0445/ |   | 28-Sep-2020 04:45
20200927-0442/ |   | 27-Sep-2020 04:42
20200926-0443/ |   | 26-Sep-2020 04:43
...

I unfortunately don't have skills to correct the script myself :-(

Thanks!

yannpub:
Thank you guys !

PR #9707, opened 2020-10-04
vlc-nightly: update url do

miccal:

- [x] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-versions/pulls) for the same update.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Closes https://github.com/Homebrew/homebrew-cask-versions/issues/9700

I have set this up to download the nightly builds of VLC version 4:

https://artifacts.videolan.org/vlc/nightly-macos/

Nightly builds of version 3 are located at:

https://artifacts.videolan.org/vlc-3.0/nightly-macos/

which would change the `base_url`.

### DIFF
--- a/Casks/vlc-nightly.rb
+++ b/Casks/vlc-nightly.rb
@@ -4,9 +4,11 @@ cask "vlc-nightly" do
 
   url do
     require "open-uri"
-    base_url = "https://nightlies.videolan.org/build/macosx-intel/last/"
-    file = URI(base_url).open.read.scan(/href="([^"]+.dmg)"/).flatten.first
-    "#{base_url}#{file}"
+    base_url = "https://artifacts.videolan.org/vlc/nightly-macos/"
+    folder = URI(base_url).read[/\d+-\d+/]
+    updated_url = "#{base_url}#{folder}/"
+    file = URI.join(updated_url).read[/href="([^"]+.dmg)"/, 1]
+    "#{updated_url}#{file}"
   end
   name "VLC media player"
   desc "Open-source cross-platform multimedia player"


Casks/vlc-nightly.rb
```diff
@@ -4,9 +4,11 @@
 
   url do
     require "open-uri"
-    base_url = "https://nightlies.videolan.org/build/macosx-intel/last/"
-    file = URI(base_url).open.read.scan(/href="([^"]+.dmg)"/).flatten.first
-    "#{base_url}#{file}"
+    base_url = "https://artifacts.videolan.org/vlc/nightly-macos/"
+    folder = URI(base_url).open.read.scan(/(\d+-\d+)/).flatten.first
+    updated_url = "#{base_url}#{folder}/"
+    file = URI.join(updated_url).open.read.scan(/href="([^"]+.dmg)"/).flatten.first
```
reitermarkus, line 10: Why `URI.join` if there is only a single string argument?

Casks/vlc-nightly.rb
```diff
@@ -4,9 +4,11 @@
 
   url do
     require "open-uri"
-    base_url = "https://nightlies.videolan.org/build/macosx-intel/last/"
-    file = URI(base_url).open.read.scan(/href="([^"]+.dmg)"/).flatten.first
-    "#{base_url}#{file}"
+    base_url = "https://artifacts.videolan.org/vlc/nightly-macos/"
+    folder = URI(base_url).open.read.scan(/(\d+-\d+)/).flatten.first
```
vitorgalvao, line 8: I think we do this for other casks as well, but there’s no point to using `scan` if we’re going to use the first result. `match` does it fine.

```suggestion
    folder = URI(base_url).open.read.match(/(\d+-\d+)/).to_s
```

miccal: Done.

Casks/vlc-nightly.rb
miccal: Because without it I get the error:
```
Error: Download failed on Cask 'vlc-nightly' with message: undefined method `call' for URI:Module
```

Casks/vlc-nightly.rb
vitorgalvao: Same thing can be done below.

miccal: Also done.

reitermarkus: Or just `.read[/(\d+-\d+)/, 1]`.

miccal: > Or just `.read[/(\d+-\d+)/, 1]`.

That is better -- makes it clear that it is the _first_ match in the list. Thanks @reitermarkus.

vitorgalvao: > Or just `.read[/(\d+-\d+)/, 1]`

What’s the `, 1` for?

miccal: > > Or just `.read[/(\d+-\d+)/, 1]`
> 
> What’s the `, 1` for?

First match.

vitorgalvao: > First match.

It’s always the first match, there’s only one capture being made. For the `1` to be meaningful, there would need to be more than one set of parenthesis. Which we can also remove, because we don’t need a capture group in this case: `URI(base_url).read[/\d+-\d+/]`

miccal: Good point -- updated.

reitermarkus (approved):

miccal:
Thank you @reitermarkus.

PR #9836, opened 2020-10-29
Add zulu15 - Azul Zulu JDK 15 (new cask)

findepi:

- [ ] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-versions/pulls) for the same update.

- [ ] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [ ] `brew cask audit --new-cask {{cask_file}}` worked successfully.
- [x] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).

### DIFF
--- a/Casks/zulu15.rb
+++ b/Casks/zulu15.rb
@@ -1,0 +1,14 @@
+cask "zulu15" do
+  version "15.0.1,15.28.13-ca"
+  sha256 "705b041d13c2499d3c6b42eacee1190ca28f144d252dc5836583365578b4ec0f"
+
+  url "https://cdn.azul.com/zulu/bin/zulu#{version.after_comma}-jdk#{version.before_comma}-macosx_x64.dmg",
+      referer: "https://www.azul.com/downloads/zulu/zulu-mac/"
+  name "Azul Zulu Java Standard Edition Development Kit"
+  desc "Azul Zulu Java Standard Edition Development Kit, version 15"
+  homepage "https://www.azul.com/downloads/zulu/zulu-mac/"
+
+  pkg "Double-Click to Install Zulu #{version.major}.pkg"
+
+  uninstall pkgutil: "com.azulsystems.zulu.#{version.major}"
+end


miccal:
Thank you @findepi, but this version of `zulu` is already available in the main [`homebrew-cask` repo](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/zulu.rb).

findepi:
@miccal thanks for taking the time to review this.
I actually didn't know about that, i must missed that when searching.
I understand a perspective where `zulu15` package becomes useful only after `zulu` package moves to version 16.
However, as a java developer, i need to know precisely which version of the JDK i am installing, so "version-less" `zulu` package that install whatever is the latest stable is not useful for me.
What are your thoughts about this?

miccal:
The latest **stable** versions are tracked in the main `homebrew-cask` repo, so if version 15 is considered the latest stable one then the `zulu` Cask points to that.

If version 16 becomes the latest stable version, then a `zulu15` Cask may be added to the `homebrew-cask-versions` repo.

findepi:
@miccal i understand what you said. That's the policy.
I am not sure, however, whether i was succesful in expressing my thoughts -- 

I am concerned that if i install `zulu` to get JDK v 15, some day it will get updated (perhaps by me doing "upgrade all there is to upgrade") to 16 -- is this part reasonable?
And, once it gets updated to 16, all the related tools will break (eg Intellij's projects, `jenv` configurations).
To me this means: because I need a specific Java version, i can install it with Homebrew only if it is _not_ the latest version released.
Does it make any sense? Or am i mixing things?

miccal:
No, it makes sense!

Homebrew, and by extension Homebrew Cask, is an evergreen package manager which means that we try to keep casks in the main repo at the latest versions, and alternate versions reside in the versions repo.

If you are really worried about this, I would suggest that you do what I do and run `brew upgrade --dry-run` to inspect what would be upgraded _before_ running `brew upgrade` so you can keep an eye on what is being upgraded, and when it happens that `zulu` is bumped to version 16 you can choose not to upgrade it and then perhaps submit version 15 to this repo.

PR #10101, opened 2020-12-16
Add arm64 specific version / checksum / url for zulu13

erinacio:

- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

~~Additionally, **if adding a new cask**:~~ Not adding a new cask, skipped.

------

I wonder if we should make `zulu13` install arm64 version on apple silicon devices, or we should create a new cask. The former one seems more intuitive for me, so I created this PR.

If this PR is approved and merged, I may create PRs for arm64 versions of `zulu11` and `zulu8` as well. Azul doesn't released arm64 version of `zulu`(15) and `zulu7` right now, so they will remain untouched.

### DIFF
--- a/Casks/zulu13.rb
+++ b/Casks/zulu13.rb
@@ -1,9 +1,19 @@
 cask "zulu13" do
-  version "13.0.5,13.35.17-ca"
-  sha256 "ce0b48f881bcb1d347b1e98a731aea34b843c2e5c9a38a93b14b364a0b10d36c"
+  if Hardware::CPU.arch == :arm64
+    version "13.0.5.1,13.35.1017-ca"
+    sha256 "c930f8475daa9878e627521d21dd7eea541759affde1763bbbb13017c52f4115"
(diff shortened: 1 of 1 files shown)


miccal:
Thank you for your contribution @erinacio, but we do not wish to add ARM-specific Casks until Hombrew itself officially supports ARM-only Mac's and https://github.com/Homebrew/brew/issues/7857 has been resolved.

erinacio:
@miccal Homebrew itself doesn't complain that ARM build is not officially supported the homebrew installer works on an apple silicon device, and issue 7857 was closed, so I think It's OK to reopen this PR since Hombrew itself officially supports ARM-only Macs now?

PR #10158, opened 2021-01-01
deadbeef-devel: fix case-sensitivity in app artifact, add description

yurikoles:

### DIFF
--- a/Casks/deadbeef-devel.rb
+++ b/Casks/deadbeef-devel.rb
@@ -5,9 +5,10 @@ cask "deadbeef-devel" do
   url "https://downloads.sourceforge.net/deadbeef/travis/osx/master/deadbeef-devel-osx-x86_64.zip",
       verified: "downloads.sourceforge.net/deadbeef/"
   name "DeaDBeeF"
+  desc "Modular audio player for UNIX-like systems"
   homepage "https://deadbeef.sourceforge.io/"
 
-  app "deadbeef.app"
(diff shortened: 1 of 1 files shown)


yurikoles:
Can we move this to main repo since app there is no stable version for yet for years?

yurikoles:
And to clarify: @homebrew/cask is there some rule on download count? A lot of casks that I used was silently removed without a commit, so, in this case an app's [repo](https://github.com/DeaDBeeF-Player/deadbeef) is notable enough: 130 forks, 56 watchers, 1k stars, but it has only 168 in last year. The app's main platform is Linux, macOS and Windows builds wasn't promoted to stable for years.

yurikoles:
How to ping a team here? Is it only allowed to existing maintainers to ping a whole team?
Ok, `sudo ping` @vitorgalvao

vitorgalvao:
> Can we move this app to main repo since there is no stable version for yet for years?

Yes. Seems like there aren’t any stable builds, so this should’ve been added there in the first place (without `-devel`).

> How to ping a team here? Is it only allowed to existing maintainers to ping a whole team?

I think I’ve asked you this before, but please don’t ping maintainers (let alone a whole group!) so soon after making a PR—let alone on a holiday—unless you’re making a critical fix. This could very well wait (even for weeks), we’d have gotten to it.

vitorgalvao (approved):

PR #10567, opened 2021-03-23
Replace open-uri with curl

Bo98:
The Homebrew/brew CI is constantly seeing flaky audit failures for `gpg-suite-nightly`. Upon further investigation, it appears that the DNS for releases.gpgtools.org contains multiple IPs, and some don't respond.

open-uri does not try another IP when one fails, but curl does. Curl also has better proxy and HTTPS support (there are known issues with open-uri certificate handling under portable-ruby). Moving to curl also aligns these network requests with what the rest of Homebrew does.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,16 +30,14 @@ For Casks of [nightlies](https://en.wikipedia.org/wiki/Daily_build), `version`, 
 
 See [this pull request for exist-db-nightly](https://github.com/Homebrew/homebrew-cask-versions/pull/3067) for an example of the procedure.
 
-Example ([exist-db-nightly.rb](https://github.com/Homebrew/homebrew-cask-versions/blob/16b3bab91ab5b9a69ef7c456441b0e0fced56516/Casks/exist-db-nightly.rb#L6#L14)):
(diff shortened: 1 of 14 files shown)


MikeMcQuaid:
Thanks for this @Bo98. I'd love a global rubocop that bans all use of `open-uri` in Homebrew/brew and all formulae/cask taps.

MikeMcQuaid (approved):

Bo98:
I think the only cases left are:

* This
* Livecheck (https://github.com/Homebrew/brew/pull/10834)
* One in homebrew-services

Bo98:
Or that's actually me searching for `require "open-uri"`. The better check is probably `URI#open` so maybe there's more.

reitermarkus:
`curl_output` is a private API, it should not be in casks.

Bo98:
Ok, I'll make sure they're not used across all other Homebrew taps and will disable the the relevant part of brew CI for now.

Bo98:
And I'll also try to make sure `utils/curl` is not auto-required. Private API shouldn't be exposed like this - we do somewhat try to limit exposure in formulae.

reitermarkus:
> not auto-required

What is required is pretty much impossible to enforce. The right solution would be to not have any global functions and instead explicitly include modules everywhere they're needed.

MikeMcQuaid:
> `curl_output` is a private API

Let's make it a public API.

PR #10620, opened 2021-03-31
Introduce new `url do` API.

reitermarkus:
```ruby
url "https://example.com" do |page|
  "#{page.url}/#{page[/(example-(?:\d+(?:\.\d+)*)\.dmg)/, 1]}"
end
```

And this pattern can be nested, as can be seen in `vlc-nightly`.


vitorgalvao:
Failing on `handbrake`.

miccal (approved):

PR #11010, opened 2021-05-20
Update vmware-fusion-tech-preview

yurikoles:
* port changes from stable
* add livechek
* sort binaries
* old homepage is dead

### DIFF
--- a/Casks/vmware-fusion-tech-preview.rb
+++ b/Casks/vmware-fusion-tech-preview.rb
@@ -4,7 +4,7 @@ cask "vmware-fusion-tech-preview" do
 
   url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-e.x.p-#{version}.dmg"
   name "VMware Fusion Tech Preview"
-  desc "Create, manage, and run virtual machinces"
+  desc "Create, manage, and run virtual machines"
   homepage "https://blogs.vmware.com/teamfusion/tech-preview"
 
   livecheck do
--- a/Casks/vmware-fusion-tech-preview.rb
+++ b/Casks/vmware-fusion-tech-preview.rb
@@ -4,7 +4,7 @@ cask "vmware-fusion-tech-preview" do
 
   url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-e.x.p-#{version}.dmg"
   name "VMware Fusion Tech Preview"
-  desc "App to run other operating systems without rebooting"
+  desc "Create, manage, and run virtual machinces"
   homepage "https://blogs.vmware.com/teamfusion/tech-preview"
 
   livecheck do
--- a/Casks/vmware-fusion-tech-preview.rb
+++ b/Casks/vmware-fusion-tech-preview.rb
@@ -8,7 +8,7 @@ cask "vmware-fusion-tech-preview" do
   homepage "https://blogs.vmware.com/teamfusion/tech-preview"
 
   livecheck do
-    url "https://bit.ly/get-fusion-tp"
+    url "http://www.vmware.com/go/get-fusion-tp"
     strategy :header_match
   end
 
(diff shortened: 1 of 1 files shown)


yurikoles:
Stable cask update is here: https://github.com/Homebrew/homebrew-cask/pull/106067

Casks/vmware-fusion-tech-preview.rb
```diff
@@ -2,16 +2,22 @@
...
-  url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-e.x.p-#{version}.dmg",
-      verified: "download3.vmware.com/"
+  url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-e.x.p-#{version}.dmg"
   name "VMware Fusion Tech Preview"
-  homepage "https://vmwarefusion.github.io/"
+  desc "App to run other operating systems without rebooting"
+  homepage "https://blogs.vmware.com/teamfusion/tech-preview"
+
+  livecheck do
+    url "https://bit.ly/get-fusion-tp"
```
vitorgalvao, line 11: We need a proper official URL, not a redirect minification which can lead anywhere.

Casks/vmware-fusion-tech-preview.rb
```diff
@@ -2,16 +2,22 @@
   version "16530630"
   sha256 "663b3d35f23541003f34ee0f5160bd04d0113703ac0a34a509b964b21a5bd5d0"
 
-  url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-e.x.p-#{version}.dmg",
-      verified: "download3.vmware.com/"
+  url "https://download3.vmware.com/software/fusion/file/VMware-Fusion-e.x.p-#{version}.dmg"
   name "VMware Fusion Tech Preview"
-  homepage "https://vmwarefusion.github.io/"
+  desc "App to run other operating systems without rebooting"
```
vitorgalvao, line 7:
```suggestion
  desc "Create, manage, and run virtual machinces"
```

vitorgalvao (changes requested):

Casks/vmware-fusion-tech-preview.rb
cho-m: I think you have a typo
`machinces` -> `machines`

vitorgalvao: Indeed. Nice catch.

Casks/vmware-fusion-tech-preview.rb
yurikoles: This URL works since last summer, I don't know a better way to livecheck this app.
Anyway, we don't download an artifact from it and similar approach is used in [deezer cask]( https://github.com/cho-m/homebrew-cask/blob/ccd2c281a5236c16124a19535b8ec309afb88638/Casks/deezer.rb#L11).

yurikoles: Old one gives 404.

vitorgalvao: > and similar approach is used in [deezer cask](https://github.com/cho-m/homebrew-cask/blob/ccd2c281a5236c16124a19535b8ec309afb88638/Casks/deezer.rb#L11).

`deezer` has a proper URL. You’re using a `bit.ly` URL here.

yurikoles: It's not me, it's they use it across all pages on their blog.

vitorgalvao: Either way. Just use the URL that expands to.

cho-m: Was about to comment that:
```
❯ curl -sI "https://bit.ly/get-fusion-tp" | grep location
location: http://www.vmware.com/go/get-fusion-tp
```

yurikoles: Yes, you are right, there is actually a proper one behind it: http://www.vmware.com/go/get-fusion-tp

yurikoles: Thanks for one-liner, @cho-m, it's much better than my stone man approach to run `curl -v <url>` and try to filter it with eyes, which I failed.

miccal (approved):

PR #11033, opened 2021-05-22
dependabot: ignore updates to `actions/stale` and `dessant/lock-threads`

Rylan12:
See https://github.com/Homebrew/brew/issues/11401

These actions should be updated by dependabot in Homebrew/.github so, to avoid extra PRs, let's ignore them here.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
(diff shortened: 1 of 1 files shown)


Rylan12:
Duplicate of #11042

Superseded by #11042 (that's going to keep being opened anyway so let's just do it there as we'll need to wait either way)

Rylan12:
Duplicate of #11042

PR #11044, opened 2021-05-24
Remove tla-plus-toolbox-nightly

Rylan12:
It seems that the URL for `tla-plus-toolbox-nightly` (https://tla.msr-inria.inria.fr/tlatoolbox/ci/products/) no longer works. This is blocking CI for non-cask-modifying PRs (e.g. #11041 and #11042).

The homepage also links to a [GitHub repo](https://github.com/tlaplus/tlaplus), but this doesn't really provide a way to download the nightly builds as there aren't downloadable artifacts for anything except the releases.

Unless there's a better way to handle this (which there may be) I think the best thing to do is just remove the Cask. I'll also note that the non-nightly version (`tla-plus-toolbox`) has pretty low install counts which makes me feel pretty good about removing the nightly version:

```console
$ brew info tla-plus-toolbox
...
==> Analytics
install: 31 (30 days), 95 (90 days), 413 (365 days)
```


cho-m:
I wonder if there is any CI flag to skip failure.

Removal should be okay based on https://formulae.brew.sh/analytics/cask-install/365d/
Summary of installs being: `0 (30 days), 5 (90 days), 22 (365 days)`

If someone really wants Cask and URL is fixed, they can always resubmit in future.

cho-m (approved):

Rylan12:
> I wonder if there is any CI flag to skip failure.

Yeah, I don't think so since it's a syntax check issue rather than an install issue. `CI-syntax-only` doesn't help as it just causes it to run on `ubuntu-latest` instead of macOS.

PR #11125, opened 2021-06-04
Provide vagrant 2.2.15

timvisher:
This is provided because 2.2.16 has a bug where it cannot be used with
RSA SHA1 ssh keys which makes it impossible to use with vagrant-aws as
well as any other platform that only supports RSA SHA1 keys. There's
been no activity from the vagrant maintainers to fix this so until
something's done about it it should be easy to install 2.2.15. I believe
this qualifies as a 'clear demonstrable need'.

https://github.com/hashicorp/vagrant/issues/12344

- [-] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-versions/search?q=is%3Aclosed&type=Issues).

  I see some issues regarding older Vagrant versions being refused. At the
  same time I think this submission should at least be considered based on
  the 'clear demonstrable need' of anyone needing to use varant to
  interact with computer providers that require the use of SHA1 RSA keys.

### DIFF
--- a/Casks/vagrant2-2-15.rb
+++ b/Casks/vagrant2-2-15.rb
@@ -1,0 +1,33 @@
+# This is provided because 2.2.16 has a bug where it cannot be used with
+# RSA SHA1 ssh keys which makes it impossible to use with vagrant-aws as
+# well as any other platform that only supports RSA SHA1 keys. There's
+# been no activity from the vagrant maintainers to fix this so until
+# something's done about it it should be easy to install 2.2.15.
+#
+# https://github.com/hashicorp/vagrant/issues/12344
+cask "vagrant2-2-15" do
+  version "2.2.15"
+  sha256 "5c2b290c4fa2371e255c56b1e96ded3d0c838d54cb7f0e8e6cf154e9f206a20e"
+
+  url "https://releases.hashicorp.com/vagrant/#{version}/vagrant_#{version}_x86_64.dmg",
+      verified: "hashicorp.com/vagrant/"
+  name "Vagrant"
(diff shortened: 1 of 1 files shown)


Casks/vagrant2-2-15.rb
```diff
@@ -0,0 +1,33 @@
+# This is provided because 2.2.16 has a bug where it cannot be used with
+# RSA SHA1 ssh keys which makes it impossible to use with vagrant-aws as
+# well as any other platform that only supports RSA SHA1 keys. There's
+# been no activity from the vagrant maintainers to fix this so until
+# something's done about it it should be easy to install 2.2.15.
+#
+# https://github.com/hashicorp/vagrant/issues/12344
+cask "vagrant2-2-15" do
```
timvisher, line 8: This particular tokenization is required by `brew audit` but it seems quite awkward. Is there any way to show it as 2.2.15?

timvisher:
There's been a response from the maintainers on the upstream issue so this may not be necessary.

timvisher:
While it's looking increasingly unlikely for this to actually be necessary I'm still curious about the name tokenization, etc. Any thoughts on that?

bevanjkay:
@timvisher This situation has generally not been a usecase for homebrew-cask-versions. Legacy versions are utilised when an application receives a major upgrade with breaking changes / different minimum software or hardware requirements. In this instance it would be recommended that you install this particular version of Vagrant manually, as this is a once off situation and the cask would not be likely to receive further updates.

timvisher:
@bevanjkay Got it. Thanks for the explanation. :)

PR #11507, opened 2021-07-27
ferdi-beta: update livecheck

bevanjkay:

### DIFF
--- a/Casks/ferdi-beta.rb
+++ b/Casks/ferdi-beta.rb
@@ -1,13 +1,13 @@
 cask "ferdi-beta" do
-  version "5.6.0"
+  version "5.6.0-beta.8"
 
   if Hardware::CPU.intel?
-    sha256 "ea211cd14be8e98b81eeadac75920d85a54f89631512bcd06f80a7c04d777137"
+    sha256 "9e9c48771cbfda22894dad3434c2e5951d6d4219093af1bf6dc700f89fdb18d1"
 
     url "https://github.com/getferdi/ferdi/releases/download/v#{version}/Ferdi-#{version}.dmg",
         verified: "github.com/getferdi/ferdi/"
   else
-    sha256 "44dd529ece76c51d5748e6c7e06026f36eda73d2915fafe5b91282b913902f70"
+    sha256 "28a4d16638589bdf69deaae48694bae3bead50bce0fd83daee1aabe9fca12d8a"
 
     url "https://github.com/getferdi/ferdi/releases/download/v#{version}/Ferdi-#{version}-arm64.dmg",
         verified: "github.com/getferdi/ferdi/"
@@ -20,6 +20,7 @@ cask "ferdi-beta" do
   livecheck do
     url :url
     strategy :git
+    regex(/beta/i)
   end
 
   auto_updates true
--- a/Casks/ferdi-beta.rb
+++ b/Casks/ferdi-beta.rb
@@ -19,8 +19,7 @@ cask "ferdi-beta" do
 
   livecheck do
     url :url
-    strategy :git
-    regex(/beta/i)
+    regex(/^v?(\d+(?:\.\d+)+[._-]beta[._-]\d+)$/i)
   end
 
   auto_updates true


Casks/ferdi-beta.rb
```diff
@@ -20,6 +20,7 @@
   livecheck do
     url :url
     strategy :git
+    regex(/beta/i)
   end
```
bevanjkay, line 24: @samford Is this the preferred method for finding a specific type of release with a :git strategy?

This same strategy should/could be used for `hyper-canary` (their release page was used instead to ensure matching the canary releases) - https://github.com/Homebrew/homebrew-cask-versions/blob/HEAD/Casks/hyper-canary.rb

samford: > Is this the preferred method for finding a specific type of release with a :git strategy? In this case we only want tags with `beta` in them, to skip any `nightly` tags.

As mentioned in the [Regex guidelines](https://docs.brew.sh/Brew-Livecheck#regex-guidelines) section of the `brew livecheck` documentation, regexes need a capture group around the version, otherwise livecheck won't work properly. The primary purpose of the regex is to extract version information from text but it's almost always used to determine what to match and what to exclude as well.

If you only want to match the `beta` tags, we should use `/^v?(\d+(?:\.\d+)+[._-]beta[._-]\d+)$/i` here, which is a modification of the standard regex for Git tags like `1.2.3`/`v1.2.3` (i.e., `/^v?(\d+(?:\.\d+)+)$/i`). This only matches the `beta` tags (not stable versions, `nightly` tags, etc.), which I assume is what you want for this cask if you're downgrading from the later `5.6.0` stable version to the beta version before it.

---

Past that, please remove `strategy :git` from this `livecheck` block. As mentioned in the [General guidelines](https://docs.brew.sh/Brew-Livecheck#general-guidelines) section of the `brew livecheck` documentation, we only use `strategy` when it's necessary (i.e., we want to use a different strategy than the one livecheck selects for a URL or we need to use a `strategy` block). In this case, livecheck already uses the `Git` strategy, so `strategy` is unnecessary.

This is one of the areas where casks don't follow the established livecheck standards (much to my chagrin). Basically, one person made this mistake (ignoring my feedback about it) and now it appears to be a persistent thing in homebrew/cask, since people continue to mimic incorrect `livecheck` blocks. I eventually plan to go through all the cask `livecheck` blocks to fix issues like these (and improve some of the patterns) but it's going to be a good amount of work and I probably won't get to it until after I take care of a few more higher-priority items on my livecheck to-do list.

---

Another side note, I've been trying to steer folks toward always having `strategy` appear as the last DSL method call in a `livecheck` block, as a way of ensuring that `regex` would appear before it. The `regex` is often passed into a `strategy` block and we want it to appear before the strategy block, so it's clear it's the same regex that's referenced in the `strategy` block arguments.

Additionally, if `regex` appears after a `strategy` block, people may think that the regex wasn't defined before being passed into the `strategy` block and this could cause some confusion. Due to how livecheck handles `strategy` blocks, the regex is defined before the `strategy` block is executed but this isn't apparent from a `livecheck` block, so it's better to structure things in a way that won't cause unnecessary confusion.

---

Just a general heads up, we only add a `livecheck` block to a formula/cask when livecheck's default behavior isn't sufficient. Unless I'm mistaken, when this `livecheck` block was originally added in #11419, livecheck was already doing the same thing by default (i.e., checking the `url` using the `Git` strategy) and the `livecheck` block shouldn't have been added.

Long story short, unnecessary `livecheck` blocks increase our maintenance burden without providing any notable benefit. When a formula/cask doesn't have a `livecheck` block, livecheck can be flexible about its behavior and we can make internal changes to livecheck and its strategies that will affect related formulae/casks without us needing to update a `livecheck` block.

bevanjkay: Thanks for the clear explanation @samford - I only became aware of unchanged `:git` blocks becoming unnecessary recently and have been trying to clean these up also when they re-appear. (Removing `strategy :page_match` where possible and now unchanged `:git` livecheck blocks).

My reasoning for only applying the `/beta/i` was to match all tags with `beta` included and then pass the array to be sorted as per usual in the `:git` strategy. But understand that it is better to be more explicit.

---

Additionally, I am happy to assist in updating the `cask` livecheck blocks when there are more solid guidelines and documentation in place, this is also why I have been trying to ask for clarification when needed.

I truly appreciate you taking the time to be clear and thorough with your answers.

PR #11626, opened 2021-08-11
firefox-esr 78.13.0

miccal:


numbermaniac:
@miccal Firefox 91.0 ESR [was also released](https://www.mozilla.org/en-US/firefox/91.0esr/releasenotes/) at the same time. Should we stick with 78.13 ESR or update the cask to 91 instead?

miccal:
> @miccal Firefox 91.0 ESR [was also released](https://www.mozilla.org/en-US/firefox/91.0esr/releasenotes/) at the same time. Should we stick with 78.13 ESR or update the cask to 91 instead?

I would suggest we update this one to 91.0 -- please feel free to open a PR @numbermaniac.

PR #11638, opened 2021-08-12
postgres-beta: remove

miccal:
I suggest we remove this in favour of the [main version](https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/postgres-unofficial.rb) as this is both unpopular and for a one-off Arm-only release.


dj110919871:
beta

carlocab (approved):

PR #11691, opened 2021-08-18
Add 1Password 8 Early Access beta channel

unitof:

### DIFF
--- a/Casks/1password-early-access.rb
+++ b/Casks/1password-early-access.rb
@@ -1,0 +1,34 @@
+cask "1password-early-access" do
+  version "8.2.0-56.BETA"
+
+  if Hardware::CPU.intel?
+    sha256 "5569970e5ea9b4fb2ed1ce2a0de226612e52e32a285c35720ce9eda68af150a1"
+    url "https://downloads.1password.com/mac/1Password-#{version}-x86_64.zip"
+  else
+    sha256 "fd647e20d5c9506cec27eba6dec6b317fb77efd7a6b17da293c89da51a415417"
+    url "https://downloads.1password.com/mac/1Password-#{version}-aarch64.zip"
+  end
+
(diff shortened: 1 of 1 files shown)


unitof:
I _think_ this is the best name, as it is more futureproof than `1password8`. Additionally, the `.app` no longer includes the major version number.

bevanjkay:
@unitof How does this differ from the existing `1password-beta` cask?
https://github.com/Homebrew/homebrew-cask-versions/blob/HEAD/Casks/1password-beta.rb

unitof:
1Password 7, the current stable major version, has its own beta channel and current beta version.

1Password 8 is a "technical preview" of a forthcoming major rewrite, with its own `nightly` and `beta` channels.

Since development on 1Password 7 (including betas) will likely continue a few weeks/months before 1Password 8 is released, I thought `beta` should stay at 7.

There's also the issue that the 7 Beta will **not** auto-update to 8, even though the formula implies it will. That could be confusing to folks who have the `1password-beta` cask installed now that it's been bumped to 8.

Would it make sense to remove `:auto-updates true` from the now-updated `1password-beta` formula?

unitof:
Which also makes the `zap` a bit difficult, since many of the app artifacts and identifiers have changed between 7 & 8, and it's not always clear which version(s) someone running `brew remove --zap 1password-beta` actually wants to zap.

bevanjkay:
@unitof regarding the zap stanza, it is stored at the time of install. So when running `brew uninstall --zap 1password-beta` it will use the zap stanza that was present when the current version of the cask was installed.

unitof:
Oh neato! Didn't know that.

Any thoughts on removing `auto-update`, since that's not true from 1Password 7 Beta -> 1Password 8 Beta?

PR #11698, opened 2021-08-19
Update dotnet-sdk-preview and dotnet-preview for arm64

ks698:

### DIFF
--- a/Casks/dotnet-preview.rb
+++ b/Casks/dotnet-preview.rb
@@ -15,8 +15,8 @@ cask "dotnet-preview" do
       end
     end
   else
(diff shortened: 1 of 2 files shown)


SommerEngineering:
Wired error message 🙁 I have checked the hash of the SDK (Preview 7, Apple Silicon). Why does the check fail? I do not see the problem.

bevanjkay:
@SommerEngineering The error is because `brew livecheck` is returning an empty string at the moment. There may be an issue with the regex, or a different problem.

bevanjkay:
@ks698 Reopened as #11702 so that we can add commits, it's better to open the PR from a branch, not your master branch.

PR #11920, opened 2021-09-11
Update affinity-publisher-beta from 1.9.2.1024 to 1.10.2.1156

unitof:


bevanjkay:
I'm not sure how well the current livecheck will work, as it looks like upstream may remove the reference to the latest version when the builds are promoted to the release branch.

PR #12406, opened 2021-11-01
Remove insomnia-alpha/inso-alpha casks

curiositycasualty:
Got word from an inso/insomnia maintainer:
> I can pen-up a PR to remove the alpha casks if you like. My hope was that the alpha casks wouldn’t require manual interventions for version/checksum updates; but that doesn’t seem to be the case.

> Yes please. Sometimes we create an alpha from PRs just for testing and there's no guarantees around them so ideally we don't have any distribution channels looking at alpha builds

So I had assumed the pre-releases which are available on github were fit for public consumption. And that turns out to not always be the case-- meaning these casks are better off removed.


curiositycasualty:
More from maintainer:
> Does this also remove the beta casks? Betas are fit for public consumption, only alphas aren't
> From within the app (for Insomnia) users can opt into either the beta or stable channels, but not alpha

Let me pen up a replacement PR for this that renames the casks, and reigns in the `page_match` to exclude `alpha`s.

PR #12926, opened 2022-01-03
Add Aerial-beta 2.3.4beta8

notDavid:
- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-versions/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [ ] `brew audit --new-cask <cask>` worked successfully.
- [x] `brew install --cask <cask>` worked successfully.
- [x] `brew uninstall --cask <cask>` worked successfully.

Hi there, so `brew audit --new-cask` gives me this error, but i don't understand why... can you advise if / how to solve this?
Thank you!

```
❯ brew audit --new-cask aerial-beta
==> Downloading https://github.com/JohnCoates/Aerial/releases/download/v2.3.4beta8/Aerial.saver.zip
==> Downloading from https://objects.githubusercontent.com/github-production-release-asset-2e65be/44998092/ae528e0b-ccc3-4391-810e-b546c
######################################################################## 100.0%
audit for aerial-beta: failed
 - Version '2.3.4beta8' differs from '2.3.3' retrieved by livecheck.
Error: 1 problem in 1 cask detected
```

PR #12930, opened 2022-01-04
aerial-beta: fix zap

miccal:
Continuation of https://github.com/Homebrew/homebrew-cask-versions/pull/12926

Apologies @notDavid -- I misunderstood what you were trying to do in your original `zap` stanza.


miccal:
@notDavid in general no, we try not to duplicate the versions in the two repositories.

miccal:
Yes it is strange, but common -- I for example use `tor-browser` and regularly switch from `tor-browser` to `tor-browser-alpha` depending on which one is "newer". I just `brew uninstall {one}` and `brew install {the other}` as required.

PR #13248, opened 2022-02-07
add kindaVim1-beta

godbout:

### DIFF
--- a/Casks/kindavim-beta.rb
+++ b/Casks/kindavim-beta.rb
@@ -1,12 +1,17 @@
 cask "kindavim-beta" do
-  version "1.0.0"
-  sha256 "4c94284b2a31f2e4959005ff0f771d65f5d4f9b69c3d5326fcd0359423eb361e"
+  version "2.0.0"
+  sha256 "0f6d6b1af1448c635fdad0c88370cd6382c21c0b34aae41daecef89f61ff760a"
 
   url "https://github.com/godbout/kindaVim.theapp/releases/download/#{version}/kindaVim.zip",
       verified: "github.com/godbout/kindaVim.theapp/"
   name "kindaVim"
   desc "Use Vim in input fields and non input fields"
   homepage "https://kindavim.app/"
+
+  livecheck do
+    url :url
+    strategy :github_latest
+  end
 
   conflicts_with cask: "kindavim"
   depends_on macos: ">= :monterey"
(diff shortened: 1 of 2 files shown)


miccal:
@godbout for the `livecheck` to work correctly here, can you please advise what versioning scheme you plan to use in the future for your beta releases -- is it going to be `1b.NUMBER` or ... ?

godbout:
@miccal apologies if this is going to be a bit specific. the stable version (in homebrew-cask) is limited. users need a license to use the app for the whole day. kindaVim1-beta doesn't need though. so the idea was to have kindaVim1-beta here, so that people who tested and were happy with the state of it can still download and don't subscribe. if i do betas in the future, they'll be including the kindaVim version number (that is not semver, it just goes up one major number each time. you subscribe and always get the latest version). current stable is v2. if i do betas of it (not planned) i would plan to add a kindaVim2-beta to homebrew-casks-versions. 

hope this makes sense? 
so kindaVim1-beta (that you unfortunately renamed without the 1, so it will make things more difficult in the future) will never be updated. so no livecheck.

miccal:
The naming follows our strict [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference), so I am afraid the name `kindaVim1` will not be accepted -- it must be `kindaVim-{something}` in this repo.

However, `beta` may not be a good fit here ...

godbout:
alright then. i saw stuff like `1password6`, `dash2`, `dash3` etc. so i thought it would be fine. in that case, and with the edit you made, it just doesn't make sense anymore for me. that would be more of a beta of the current version, rather than giving a way for users to get the old beta through homebrew. gonna close the PR then. they'll have to download through GitHub. i'll add back a kV-beta if one day there's a real, current, live one. thanks.

miccal:
Those are for specific versions -- for example `dash2` install dash version 2.x.x as the version of dash in the main repo is a higher version.

So if `kindaVim1` installed version 1.x.x of kindaVim, then we could indeed call it just `kindaVim1`.

godbout:
yeah that was exactly the idea, but at the same time kV1-beta will never get updated. it was to give a way for users to download/install kV1-beta (without subscription) through their Brew files. updates will never happen. so i guess ultimately it's better not to pollute cask-versions. they'll just have to download through GitHub instead. thanks for your time, and for ultimately making things clearer and better.

miccal:
You're welcome @godbout.

PR #13424, opened 2022-03-02
google-chrome-canary: use latest version

hannesbraun:
I think we can switch to `:latest` for the version here since the exact version number isn't required for the cask to work. Since there are sometimes multiple updates on a single day, this should contribute to lower the maintenance work for the repository.

If this get's merged, I'd update `google-chrome-beta` and `google-chrome-dev` in the same way.

---


miccal:
Thanks @hannesbraun but I would prefer to keep these versioned to match the `google-chrome` Cask in the main repository.

PR #13476, opened 2022-03-10
scribus-dev: remove in favor of scribus in homebrew-cask

neutric:
The version of Scribus in `homebrew-cask` is 32-bit only and thus doesn't run on Catalina and newer. It should thus considered as the "[But There Is No Stable Version!](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version)" case which makes a development version in `homebrew-cask` acceptable. Please merge together with https://github.com/Homebrew/homebrew-cask/pull/120289/.


neutric:
Oops, in an effort to clean up old cruft that was waiting for my passsword since last night, which I just gave it, I managed to delete my fork, which closed this. Will reopen momentarily.

PR #13705, opened 2022-04-13
Add kitty-nightly

timbedard:

### DIFF
--- a/Casks/kitty-nightly.rb
+++ b/Casks/kitty-nightly.rb
@@ -7,17 +7,18 @@ cask "kitty-nightly" do
   desc "GPU-based terminal emulator"
   homepage "https://github.com/kovidgoyal/kitty"
 
+  conflicts_with cask: "kitty"
   depends_on macos: ">= :sierra"
 
-  app "kitty.app", target: "kitty nightly.app"
+  app "kitty.app"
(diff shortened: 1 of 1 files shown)


Casks/kitty-nightly.rb
```diff
@@ -0,0 +1,31 @@
...
+  name "kitty-nightly"
+  desc "GPU-based terminal emulator"
+  homepage "https://github.com/kovidgoyal/kitty"
+
+  depends_on macos: ">= :sierra"
+
+  app "kitty.app", target: "kitty nightly.app"
+  # shim script (https://github.com/Homebrew/homebrew-cask/issues/18809)
+  shimscript = "#{staged_path}/kitty.wrapper.sh"
+  binary shimscript, target: "kitty-nightly"
```
bevanjkay, line 15: @timbedard Rather than renaming this to `kitty-nightly` it would be better to leave it the same as the main cask, but add a `conflicts_with` stanza so that they cannot be installed at the same time. Looking at the zap it looks like they would share preference files etc.. so should only install one version at a time.

bevanjkay (changes requested):

bevanjkay (approved):

PR #13818, opened 2022-05-01
Pass regex into strategy block, use #scan

samford:

- [ ] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [ ] `brew install --cask <cask>` worked successfully.
- [ ] `brew uninstall --cask <cask>` worked successfully.

---

The existing `livecheck` blocks in this PR contain a `strategy` block that uses a regex internally. This PR updates these `livecheck` blocks to define the regex using `#regex` and simply pass the regex into the `strategy` block. This approach is preferable, as it aligns with how most `livecheck` blocks work, makes it easy to add/remove a `strategy` block in the future (i.e., the regex doesn't have to be manually moved into/out of the `strategy` block), allows the regex value to be surfaced in the `brew livecheck` debug and JSON output, and allows us to audit the regex.

[This approach can only be used with a single regex at the moment but I'll be creating a brew PR in the near future to allow `#regex` to handle multiple regexes. Once this is done, we can standardize on this approach for all related `strategy` blocks. In the interim time, I'm cleaning up `strategy` blocks that use one regex internally (starting with the smaller cask repositories). This is partly related to my ongoing work to address livecheck RuboCop offenses, as these changes resolve the `A regex is required if strategy :page_match is present` audit.]

---

This PR also updates some of these `strategy` blocks to use the `page.scan(regex).map { |match| ... }` approach instead of the multiline `match = page.match(regex)`/`next if match.blank?`/`"#{match[1]},#{match[2]}"` approach. The `#scan` approach is generally preferred because it will handle all matches in the content, whereas `#match` will only handle the first match (though in contexts where the content will only ever contain one match, we use `[regex, 1]` if the regex only has one capture group and the `#match` approach for more than one capture group`). Besides that, the `#scan` approach is generally less verbose.

One thing to note about converting from `#match` to `#scan` is that the capture group index numbers have to be decremented by one. Basically, the first capture group from `#scan` is index 0 but the first capture group index for `#match` is 1 instead (index 0 is the full matched text).

---

There are some other casks in this tap that need to be updated like this but the changes were a little less simple. I'll be handling these in separate PRs, where I can explain the changes.

### DIFF
--- a/Casks/arduino-ide-beta.rb
+++ b/Casks/arduino-ide-beta.rb
@@ -9,11 +9,9 @@ cask "arduino-ide-beta" do
 
   livecheck do
     url "https://www.arduino.cc/en/software/"
-    strategy :page_match do |page|
-      match = page.match(/href=.*?arduino[._-]ide[._-]v?(\d+(?:\.\d+)+)[._-]beta\.(\d+)[._-]macos[._-]64bit\.dmg/i)
-      next if match.blank?
-
-      "#{match[1]},#{match[2]}"
+    regex(/href=.*?arduino[._-]ide[._-]v?(\d+(?:\.\d+)+)[._-]beta[._-]?(\d+)[._-]macos[._-]64bit\.dmg/i)
+    strategy :page_match do |page, regex|
+      page.scan(regex).map { |match| "#{match[0]},#{match[1]}" }
     end
   end
 
--- a/Casks/background-music-pre.rb
+++ b/Casks/background-music-pre.rb
@@ -7,10 +7,14 @@ cask "background-music-pre" do
   desc "Audio utility"
   homepage "https://github.com/kyleneideck/BackgroundMusic"
 
+  # When the leading version for two snapshot tags are the same, we can't
+  # determine which is newer based on the trailing hash. As such, this check
+  # treats the most recent snapshot tag as newest (if any).
   livecheck do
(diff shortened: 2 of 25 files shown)


samford:
CI is predictably failing in the `brew audit --cask --tap homebrew/cask-versions` step, because `sequel-pro-nightly` has a URL that's defined at runtime by fetching/parsing an XML file but the upstream server has an expired SSL certificate:

```
audit for sequel-pro-nightly: failed
 - exception while auditing sequel-pro-nightly: Failure while executing; `/usr/local/Homebrew/Library/Homebrew/shims/shared/curl --disable --cookie /dev/null --globoff --show-error --user-agent Homebrew/3.4.9-33-g3088bbf\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 10.15.7\)\ curl/7.64.1 --header Accept-Language:\ en --retry 3 --fail --silent --location https://sequelpro.com/builds/latest-test-build.xml` exited with 60. Here's the output:
curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

This PR doesn't modify the `sequel-pro-nightly` cask, so it's unrelated.

PR #14289, opened 2022-07-13
Remove alfred5, add alfred4

p-linnane:
Alfred 5 has been released

https://www.alfredapp.com/alfred-5-whats-new/


bevanjkay:
@p-linnane We may be best to leave the livecheck for now as the version 4 branch may still receive bug fixes, unless you have seen an announcement that it won't?

p-linnane:
@bevanjkay I support that idea, and have readied it. I just modeled it after v2 and v3, but obviously those have been EOL much longer.

PR #14411, opened 2022-08-01
Update temurin11 to latest release

dhalperi:
For reasons not yet clear to me, they released updated binaries for x64 but not aarch64.

See: https://adoptium.net/temurin/releases?version=11

and: https://github.com/adoptium/temurin11-binaries/releases/tag/jdk-11.0.16%2B8
vs: https://github.com/adoptium/temurin11-binaries/releases/tag/jdk-11.0.15%2B10

FYI, I ran the commands above only on an x64 macOS laptop. I do not currently have access to an aarch64 laptop.

### DIFF
--- a/Casks/temurin11.rb
+++ b/Casks/temurin11.rb
@@ -1,11 +1,11 @@
 cask "temurin11" do
   arch = Hardware::CPU.intel? ? "x64" : "aarch64"
 
-  version "11.0.15,10"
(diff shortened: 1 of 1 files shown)


gdams (approved):
Approved - the aarch64 binary will be available shortly but let’s proceed with this to unblock the intel users

PR #14845, opened 2022-10-14
Add Yubico Authenticator Beta

joshka:

- [X] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [X] `brew audit --cask --online <cask>` is error-free.

- [X] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [X] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-versions/search?q=is%3Aclosed&type=Issues).
- [X] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).

### DIFF
--- a/Casks/yubico-authenticator-beta.rb
+++ b/Casks/yubico-authenticator-beta.rb
@@ -12,7 +12,6 @@ cask "yubico-authenticator-beta" do
     regex(/yubioath[._-]desktop[._-]v?(\d+(?:\.\d+)*-beta(?:\.\d+)*)[._-]mac\.dmg/i)
   end
 
-  conflicts_with cask: "yubico-authenticator"
-
-  app "Yubico Authenticator.app"
+  # Renamed to avoid conflict with yubico-authenticator.
+  app "Yubico Authenticator.app", target: "Yubico Authenticator Beta.app"
 end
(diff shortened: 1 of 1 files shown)


joshka:
Note: https://github.com/Homebrew/homebrew-cask-drivers/blob/master/Casks/yubico-authenticator.rb has some an uninstall and zap stanza. Checking the installation after running the beta version (concurrently with the stable version) it seems none of the listed files are actually created. Hence I have not added them here.

miccal (approved):

miccal:
Thank you @joshka.

joshka:
@miccal as an aside, I have both the beta and stable version of this installed - there's no conflict per say other than the name of the application. I see you settled on the rename - is this the normal way, or does `-versions` tend to go with the conflicts approach?

miccal:
We would usually go with the `conflicts_with` approach, but it a a little flaky between the `drivers` and `versions` repositories, so I went with your original approach.

PR #14969, opened 2022-10-27
virtualbox-beta: update and add arm64 version

yurikoles:

---
I originally submitted it for a stable version, but https://github.com/Homebrew/homebrew-cask/pull/134443 was rejected.

### DIFF
--- a/Casks/virtualbox-beta.rb
+++ b/Casks/virtualbox-beta.rb
@@ -3,14 +3,29 @@ cask "virtualbox-beta" do
     version "7.0.0_BETA3,153829"
     sha256 "a97ad4e37f975ec3ec093a1dfc58f456cac2066f27ccc743a523a261235785b0"
     url "https://download.virtualbox.org/virtualbox/#{version.csv.first}/VirtualBox-#{version.csv.first}-#{version.csv.second}-OSX.dmg"
+
+    livecheck do
+      url :homepage
+      regex(/href=.*?VirtualBox[._-]v?(\d+(?:\.\d+)+[._-][^._-]+?)[._-](\d+)[._-]OSX\.dmg/i)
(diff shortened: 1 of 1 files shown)


samford (commented):
I was hoping that we could check the ["Testbuilds" page](https://www.virtualbox.org/wiki/Testbuilds) (i.e., the homepage) but the URLs are notably different. Those files are found in https://www.virtualbox.org/download/testcase/ (instead of https://download.virtualbox.org/virtualbox/) and the current snapshot revision (153872) differs from revision in the newest unstable directory (153829 for 7.0.0_BETA3). There isn't a link for an ARM dmg file anyway.

I've added a `livecheck` block for Intel but the situation for ARM is a bit weird at the moment. The newest unstable version with a directory is 7.0.0_BETA3 but the ARM dmg files were created after that and were simply added to corresponding stable version directories (7.0.0 for 7.0.0_BETA4-153978 and 7.0.2 for 7.0.2_BETA4-154219) instead of creating new unstable version directories.

In the current state, there's a chance that unstable ARM versions will continue to appear in stable version directories but there's also a chance they may appear in an unstable directory the next time one is created. Our only option at the moment would be to fetch https://download.virtualbox.org/virtualbox/, identify the latest unstable and stable directories, fetch the pages for both directories and try to match ARM versions on each, and return the newest unstable version from either. This would require three requests (not great), so I've added a `TODO` comment in the `on_arm` block for us to revisit this if/when the upstream situation becomes more clear/predictable.

That said, I can technically implement a lengthy `livecheck` block for ARM if we really want one but I'm not very keen on having livecheck dig around multiple pages trying to find a version simply because upstream is doing something weird.

miccal (approved):

Issue #14993, opened 2022-10-29
Migrate all casks from `appcast` to `livecheck` & review existing `livecheck` blocks


p-linnane:
### Description of issue

A large number of casks in `homebrew-cask-versions` are still using `appcast` instead of `livecheck`. I'm opening this issue as a tracking list for working through the migration. 

I will also be verifying existing `livecheck` blocks and marking old versions as `discontinued` if they have not been updated in a significant amount of time.

This will ensure things are cleaner for the eventual `homebrew-cask` repo consolidation discussed in https://github.com/Homebrew/homebrew-cask/issues/112102.

- [x] 1password-beta (Existing livecheck ✅)
- [x] 1password-cli-beta (Existing livecheck ✅)
- [x] 1password-cli1 (Existing livecheck ✅)
- [x] 1password6 #14994
- [x] 1password7 (Existing livecheck ✅)
- [x] ableton-live-intro10 #14995
- [x] ableton-live-lite10 #14996
- [x] ableton-live-standard10 #14997
- [x] ableton-live-suite10 #14998
- [x] ableton-live-suite9 #14999
- [x] adguard-nightly #15000
- [x] adguard-vpn-nightly #15001
- [x] adoptopenjdk8 (Already discontinued ✅)
- [x] aerial-beta #15002
- [x] affinity-designer-beta #15003
- [x] affinity-photo-beta #15004
- [x] affinity-publisher-beta #15005
- [x] alfred2 #15006
- [x] alfred3 #15007
- [x] alfred4 (Existing livecheck ✅)
- [x] anaconda2 #15008
- [x] android-studio-preview-beta (Existing livecheck ✅)
- [x] android-studio-preview-canary (Existing livecheck ✅)
- [x] anki-beta #15009
- [x] appcode-eap (Existing livecheck ✅)
- [x] arduino-ide-nightly (`:latest`, no reliable version info ✅)
- [x] arduino-nightly (Already discontinued ✅)
- [x] arq5 #15010
- [x] atom-beta #15011
- [x] atom-nightly #15012
- [x] azure-data-studio-insiders #15013
- [x] background-music-pre (Existing livecheck ✅)
- [x] blender-lts (Existing livecheck ✅)
- [x] brave-browser-beta (Existing livecheck ✅)
- [x] brave-browser-dev (Existing livecheck ✅)
- [x] brave-browser-nightly #15014
- [x] buckets-beta (Existing livecheck ✅)
[...]

PR #15049, opened 2022-10-31
lilypond-dev: Remove cask

p-linnane:
~Created with `brew bump-cask-pr`.~

https://github.com/Homebrew/homebrew-cask-versions/issues/14993


bevanjkay:
@p-linnane It looks like this doesn't ship as an `.app` anymore, just CLI. We can probably remove it as the main has also been moved to `homebrew-core` earlier this year.

p-linnane:
Shouldn't have gone to bed while CI was still running 🤦‍♂️

miccal (approved):

PR #15557, opened 2022-12-28
safari-technology-preview: convert to `on_system` blocks

miccal:


Casks/safari-technology-preview.rb
```diff
@@ -1,8 +1,9 @@
 cask "safari-technology-preview" do
-  if MacOS.version <= :monterey
+  on_monterey :or_older do
```
bevanjkay, line 2: @miccal We still need to pin these to `or_older`/`or_newer` at each end so that the casks can be read on older system, otherwise they can fail because the version stanza is missing. Because of the `depends_on` they will through the appropriate error that it isn't compatible with the system on older MacOS.

bevanjkay (approved):

PR #15770, opened 2023-01-23
Create folding-at-home-beta 8.1.11

kbernhagen:

### DIFF
--- a/Casks/folding-at-home-beta.rb
+++ b/Casks/folding-at-home-beta.rb
@@ -19,16 +19,4 @@ cask "folding-at-home-beta" do
 
   uninstall launchctl: "org.foldingathome.fahclient",
             pkgutil:   "org.foldingathome.fahclient.*"
-
-  caveats do
-    <<~EOS
-
-      Please open the Web Control app https://app.foldingathome.org in a browser other than Safari.
-
-      Safari will not allow unencrypted connections from an https page.
-
-      Install will fail if Safari seems to be your only browser.
-
-    EOS
-  end
 end
--- a/Casks/folding-at-home-beta.rb
+++ b/Casks/folding-at-home-beta.rb
@@ -18,5 +18,7 @@ cask "folding-at-home-beta" do
   pkg "fah-client_#{version}_universal.pkg"
 
   uninstall launchctl: "org.foldingathome.fahclient",
-            pkgutil:   "org.foldingathome.fahclient.*"
+            pkgutil:   "org.foldingathome.fahclient.pkg"
+
+  zap trash: "/Library/Application Support/FAHClient"
 end
--- a/Casks/folding-at-home-beta.rb
+++ b/Casks/folding-at-home-beta.rb
@@ -14,19 +14,16 @@ cask "folding-at-home-beta" do
 
   conflicts_with cask: "folding-at-home"
   depends_on macos: ">= :high_sierra"
-  depends_on arch: [:arm64, :x86_64]
 
   pkg "fah-client_#{version}_universal.pkg"
 
   uninstall launchctl: "org.foldingathome.fahclient",
             pkgutil:   "org.foldingathome.fahclient.*"
 
-  zap trash: "/Library/Application Support/FAHClient"
-
   caveats do
     <<~EOS
 
-      Please open https://app.foldingathome.org in a browser other than Safari.
+      Please open the Web Control app https://app.foldingathome.org in a browser other than Safari.
 
       Safari will not allow unencrypted connections from an https page.
 
--- a/Casks/folding-at-home-beta.rb
+++ b/Casks/folding-at-home-beta.rb
@@ -1,0 +1,37 @@
+cask "folding-at-home-beta" do
+  version "8.1.11"
+  sha256 "f6375dea6c03dd9c96c9d901e073fc2f14852866a1cab84bd4af6135283069ae"
+
+  url "https://download.foldingathome.org/releases/beta/fah-client/macos-12-universal/release/fah-client_#{version}_universal.pkg"
+  name "Folding@home Client Beta"
+  desc "Protein folding simulation for scientific research"
+  homepage "https://foldingathome.org/"
+
+  livecheck do
+    url "https://download.foldingathome.org/releases/beta/fah-client/macos-12-universal/release/"
+    regex(/href=.*?fah[._-]?client[._-]v?(\d+(?:\.\d+)*)(_universal)?\.pkg/i)
+  end
+
+  conflicts_with cask: "folding-at-home"
+  depends_on macos: ">= :high_sierra"
+  depends_on arch: [:arm64, :x86_64]
+
+  pkg "fah-client_#{version}_universal.pkg"
+
+  uninstall launchctl: "org.foldingathome.fahclient",
+            pkgutil:   "org.foldingathome.fahclient.*"
+
+  zap trash: "/Library/Application Support/FAHClient"
+
+  caveats do
+    <<~EOS
+
+      Please open https://app.foldingathome.org in a browser other than Safari.
+
+      Safari will not allow unencrypted connections from an https page.
+
+      Install will fail if Safari seems to be your only browser.
+
+    EOS
+  end
+end


kbernhagen:
I note that audit --new-cask fails iff the cask is currently installed. Is this normal?

Folding@home v8.1.11 is in beta tomorrow, and has been alpha for a week.

The stable release version is 7.6.21, which has its own (conflicting) cask.

kbernhagen:
Uninstall zap, which I copied from folding-at-home, fails with permissions errors.
I don't know how to correct this.
It looks like file owner was changed from `nobody` to myself.

Example:
```
chmod: Unable to change file mode on /Library/Application Support/FAHClient/config.xml: Operation not permitted
```

Casks/folding-at-home-beta.rb
```diff
@@ -0,0 +1,37 @@
...
+    <<~EOS
+
+      Please open https://app.foldingathome.org in a browser other than Safari.
+
+      Safari will not allow unencrypted connections from an https page.
+
+      Install will fail if Safari seems to be your only browser.
+
+    EOS
+  end
```
bevanjkay, line 36: I can't see that this should affect the cask install in any way?
This should be removed as it does not affect the operation of the cask, but only accessing the homepage.

Casks/folding-at-home-beta.rb
```diff
@@ -0,0 +1,37 @@
...
+  homepage "https://foldingathome.org/"
+
+  livecheck do
+    url "https://download.foldingathome.org/releases/beta/fah-client/macos-12-universal/release/"
+    regex(/href=.*?fah[._-]?client[._-]v?(\d+(?:\.\d+)*)(_universal)?\.pkg/i)
+  end
+
+  conflicts_with cask: "folding-at-home"
+  depends_on macos: ">= :high_sierra"
+  depends_on arch: [:arm64, :x86_64]
```
bevanjkay, line 17: This line can be ommitted.

Casks/folding-at-home-beta.rb
kbernhagen: The installer will deliberately fail if the user’s default browser is not Safari and a known compatible browser is not found in /Applications. 

A user cannot configure or control the client without a compatible browser. Folding can start and use all cpu cores with no obvious way to stop it. 

It is warning that I strongly believe should remain.

Casks/folding-at-home-beta.rb
kbernhagen: Which line?

Casks/folding-at-home-beta.rb
bevanjkay: In the case I would suggest that this is not suitable to be a cask. All installs should be headless, and the caveat will only print to console, not stop an install.
So if there is a chance of the installing hanging and using all CPU cores then the installer should be changed upstream.

Casks/folding-at-home-beta.rb
bevanjkay: `  depends_on arch: [:arm64, :x86_64]`

Casks/folding-at-home-beta.rb
kbernhagen: That is not a home page. It is a control web app.

kbernhagen: No. Like I said, the install will fall/abort.

kbernhagen: As to headless, the client can be configured for remote access. But that currently requires manual configuration. Other browsers can be brew installed to allow this install to succeed.

bevanjkay: Sorry, by headless I meant that the install should succeed without user intervention. Which it looks like in this case it does - as long as a compatible browser is found.

I have some reservations with adding this due to the possibility of the install failing for some users, perhaps upstream should have the warning occur after installation when the app is run instead. I'll see what the other maintainers think here.

Casks/folding-at-home-beta.rb
kbernhagen: Done

kbernhagen:
I removed the broken zap stanza.
If you know why it fails, please let me know.

Casks/folding-at-home-beta.rb
miccal: The `pkg` contains two scripts the are relevant here: one is called `check-not-just-safari`:
```
#!/bin/bash

# exit 0 if see a browser not Safari

cd "$(dirname "$0")"

SUDO=""
if [ "$SUDO_USER" != "" ]; then
  SUDO="sudo -u $SUDO_USER"
elif [ "$USER" != "" ]; then
  SUDO="sudo -u $USER"
fi
BROWSER=$($SUDO ./defaultbrowser.sh) || BROWSER=""
: "${BROWSER}"; printf -v BROWSER '%b' "${_//%/\\x}"

# default browser is not Safari; assume compatible
[ ! -z "$BROWSER" ] && [ "$BROWSER" != "Safari" ] && exit 0

APPS=("Google Chrome" "Firefox" "Microsoft Edge" "Brave Browser")
for APP in "${APPS[@]}"; do
  [ -d "/Applications/${APP}.app" ] && exit 0
done

exit 1
```
and the other is called `defaultbrowser.sh`:
```
#!/bin/bash
# https://discussions.apple.com/thread/251602703

# return url-encoded default browser name that would open the URL

function default_browser () {
    osascript <<-AS
    use framework "AppKit"
    use AppleScript version "2.4"
    use scripting additions

    property NSWorkspace : a reference to current application's NSWorkspace
    property NSURL : a reference to current application's NSURL

    set wurl to NSURL's URLWithString:"https://www.apple.com"
    set thisBrowser to (NSWorkspace's sharedWorkspace)'s ¬
                        URLForApplicationToOpenURL:wurl
    set appname to (thisBrowser's absoluteString)'s lastPathComponent()'s ¬
                    stringByDeletingPathExtension() as text
    return appname as text
AS
    return
}

printf '%s\n' $(default_browser)
exit 0
```
So, in my opinion the `pkg` handles this itself and hence the `caveat` is not necessary.

kbernhagen: I intended it to be informational.
If the install fails, the user gets no other feedback for why it fails.
If it succeeds, the user is pointed to the Web Control app url.

p-linnane: Just getting caught up here. If this can install without user intervention as @bevanjkay said, then it should be fine to add. 

`caveat` stanzas should be used for things that are specific to Homebrew. In this case, it is unnecessary and shoult not be included. [`homebrew-cask` is not a discoverability service](https://docs.brew.sh/Acceptable-Casks#homebrew-cask-is-not-a-discoverability-service), so users installing this software should already be aware of the idiosyncrasies that may be encountered.

We will need a working `zap` stanza in order to accept this though.

kbernhagen: Re-reading the docs, it is my understanding that `zap` is only for user-specific files. FAH version 8 does not create any.

kbernhagen: `caveats` removed

kbernhagen:
That `zap trash:` that was committed is what did not work.

miccal:
Uninstalling with `zap` often requires `sudo`, so this should be fine.

Thanks @kbernhagen.

miccal (approved):

kbernhagen:
So, I would just use this?
```
sudo brew uninstall --cask folding-at-home-beta
```
This is an exception to the sudo is bad rule?

miccal:
No, it should ask for your password when you run `brew uninstall --zap folding-at-home-beta`.

kbernhagen:
I am asked for my password by `brew uninstall --zap folding-at-home-beta`.
But the `zap` fails.
Looks like this (I have a couple thousand files):
```
% brew uninstall --zap folding-at-home-beta
==> Implied `brew uninstall --cask folding-at-home-beta`
==> Removing launchctl service org.foldingathome.fahclient
Password:
==> Uninstalling packages; your password may be necessary:
org.foldingathome.fahclient.pkg
==> Dispatching zap stanza
==> Trashing files:
/Library/Application Support/FAHClient
chmod: Unable to change file mode on /Library/Application Support/FAHClient/config.xml: Operation not permitted
chmod: Unable to change file mode on /Library/Application Support/FAHClient/log.txt: Operation not permitted

snip

chmod: Unable to change file mode on /Library/Application Support/FAHClient/client.db: Operation not permitted
==> Using sudo to gain ownership of path '/Library/Application Support/FAHClient
Warning: The following files could not be trashed, please do so manually:
/Library/Application Support/FAHClient
==> Removing all staged versions of Cask 'folding-at-home-beta'
```

kbernhagen:
Afterward, all the files' ownership changed from `nobody:nobody` to `kevin2:nobody`.
But nothing is trashed or deleted.

kbernhagen:
Could this be a brew zap bug?

PR #15778, opened 2023-01-24
whatsapp-alpha: new Cask

miccal:
Not too sure about the name `whatsapp-native`?


p-linnane:
>The WhatsApp desktop app native to Mac operating systems is currently in the early stages of development.

Assuming this will eventually replace the existing app. Maybe we want to call it `whatsapp-alpha` so that as it moves through the release cycle we can update the beta cask and then stable?

p-linnane (approved):

PR #16004, opened 2023-02-18
microsoft-office-2016: update `conflicts_with`

miccal:


bevanjkay:
@miccal Ahh there you go. I think it's required for arrays with 4 or greater items.

PR #16307, opened 2023-03-27
Add typora-dev

stevenlele:
The `zh-Hans-CN` block is to use official mirror for Chinese users.

---

### DIFF
--- a/Casks/typora-dev.rb
+++ b/Casks/typora-dev.rb
@@ -2,7 +2,7 @@ cask "typora-dev" do
   version "1.5.4"
   sha256 "3a0e36a87eb3f6441a0a822e85406d5d2a582bac7d0a01be4356bc93bb677742"
 
-  language "zh", "CN", "Hans" do # use official Chinese mirror
+  language "zh-Hans-CN" do # use official Chinese mirror
     url "https://download2.typoraio.cn/mac/Typora-#{version}-dev.dmg",
         verified: "typoraio.cn/"
   end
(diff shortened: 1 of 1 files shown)


Casks/typora-dev.rb
```diff
@@ -0,0 +1,38 @@
+cask "typora-dev" do
+  version "1.5.4"
+  sha256 "3a0e36a87eb3f6441a0a822e85406d5d2a582bac7d0a01be4356bc93bb677742"
```
razvanazamfirei, line 3: Do both casks have the same sha256?

Casks/typora-dev.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  language "zh-Hans-CN" do
```
razvanazamfirei, line 5: Any advantage to using this as opposed to a broader language code? See:
https://docs.brew.sh/Cask-Cookbook#stanza-language

razvanazamfirei: See my latest push! The question is essentially whether we should only use `zh-Hans-CN` for the mirror, or if specifying `zh`, `zh-CN`, `zh-Hans` or `CN` should also lead to the mirror being used. The previous code, `zh-Hans-CN`, will not match other zh-related codes.

If there's a rationale behind not matching other zh codes let me know!

razvanazamfirei: Awesome, thank you for clarifying!

razvanazamfirei (approved):

PR #16339, opened 2023-04-02
Add Godot LTS version

haythamnikolaidis:
Godot is currently available in [homebrew-cask](https://github.com/Homebrew/homebrew-cask) however that installs version 4 which is the latest stable version, which might not be suitable for production. However the LTS version 3.5 can no longer be installed via brew, this cask makes the LTS version available on homebrew again.

***

### DIFF
--- a/Casks/godot3.rb
+++ b/Casks/godot3.rb
@@ -13,6 +13,8 @@ cask "godot3" do
     regex(/^v?(3(?:\.\d+)+)[._-]stable$/i)
   end
 
+  conflicts_with cask: "godot"
+
   app "Godot.app"
   binary "#{appdir}/Godot.app/Contents/MacOS/Godot", target: "godot"
 
--- a/Casks/godot3.rb
+++ b/Casks/godot3.rb
@@ -1,0 +1,26 @@
+cask "godot3" do
+  version "3.5.2"
+  sha256 "2a010b8fbf8241a20224c14ecfac01f4d364d68e14e477459c61ca40716bc0a9"
+
+  url "https://downloads.tuxfamily.org/godotengine/#{version}/Godot_v#{version}-stable_osx.universal.zip",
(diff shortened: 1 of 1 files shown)


razvanazamfirei:
Does this conflict with Godot `4.x`?

p-linnane:
Given that it installs a "Godot" app, they would conflict with each other.

Casks/godot3.rb
```diff
@@ -0,0 +1,26 @@
...
+  desc "Game development engine"
+  homepage "https://godotengine.org/"
+
+  livecheck do
+    url "https://github.com/godotengine/godot"
+    regex(/^v?(3(?:\.\d+)+)[._-]stable$/i)
+  end
+
+  app "Godot.app"
+  binary "#{appdir}/Godot.app/Contents/MacOS/Godot", target: "godot3"
```
bevanjkay, line 17: Because this will conflict with the main `godot` cask, I don't think renaming the binary should be necessary.

haythamnikolaidis: Thanks, I'll remove the version from the binary. 

This is my first cask, so I am not sure if I should find a way to not conflict with Godot `4.x`. That being said as a Godot user I do not have a reason to have both versions on my machine at least not until Godot 4.1 is released.

bevanjkay: @haythamnikolaidis The cask should have a [`conflicts_with`](https://docs.brew.sh/Cask-Cookbook#stanza-conflicts_with) stanza added so that a user cannot have both this and `godot` installed at the same time.

haythamnikolaidis: Thank you for taking the time to point these things out, @bevanjkay I have added the stanza. Hopefully, I didn't miss anything else.

razvanazamfirei:
@haythamnikolaidis, thank you for your contribution to Homebrew!

razvanazamfirei (approved):

PR #14654, opened 2022-09-09
Update blender-lts from 2.93.10 to 3.3.0

wavetro:
Created with `brew bump-cask-pr`.

PR #17056, opened 2023-07-02
blender-lts: 3.3.8 to 3.6.0

wavetro:

- [ ] `brew audit --cask --online <cask>` is error-free.*

*I'm not sure how to fix this, as the existing 3.3.8 cask had the same errors:

```
audit for blender-lts: failed
 - Version '3.6.0' differs from '3.3.8' retrieved by livecheck.
 - Version '3.6.0' differs from '3.3.8' retrieved by livecheck.
homebrew/core
  * audit_exceptions/versioned_dependencies_conflicts_allowlist.json references
    formulae or casks that are not found in the homebrew/core tap.
    Invalid formulae or casks: predictionio
  * pypi_formula_mappings.json references
    formulae or casks that are not found in the homebrew/core tap.
    Invalid formulae or casks: rtv
blender-lts
  * Version '3.6.0' differs from '3.3.8' retrieved by livecheck.
Error: 3 problems in 1 cask and 1 tap detected.
```

### DIFF
--- a/Casks/blender-lts.rb
+++ b/Casks/blender-lts.rb
@@ -22,7 +22,7 @@ cask "blender-lts" do
       next if lts_page[:content].blank?
 
       lts_versions =
-        lts_page[:content].scan(%r{href=["'].*/download/lts/(\d+(?:[.-]\d+)+)/["' >]}i)
+        lts_page[:content].scan(%r{href=["'].*/download/(?:lts|releases)/(\d+(?:[.-]\d+)+)/["' >]}i)
                           .flatten
                           .uniq
                           .map { |v| Version.new(v) }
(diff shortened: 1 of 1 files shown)


wavetro:
(see #14654 for related discussion on how Blender handles LTS versions that are also the latest non-LTS stable)

samford:
The approach in the existing `livecheck` block technically still works but the difference is that the 3.6 LTS link on the `/download/lts/` page is `/download/releases/3-6/` (since 3.6 LTS is also the latest release overall). The regex in the `strategy` block only matches URLs like `/download/lts/3-3/`, so we can simply update it to match `lts|releases` (as the version links on the `/download/lts/` page will presumably all be LTS versions).

I've pushed a commit that updates the `livecheck` block accordingly. As before, we may have to update this again in the future (the upstream situation isn't ideal) but it works for now.

miccal (approved):

PR #17271, opened 2023-07-28
Add Threema Multi Device Beta

joelfischerr:
Threema is a paid cross-platform encrypted instant messaging app developed by Threema GmbH in Switzerland. The multi device beta introduces a new multi device functionality to the existing desktop app. I have included multi device in the cask name to clearly distinguish it from the previous desktop application which would also be available as a beta version.

### DIFF
--- a/Casks/threema-multi-device-beta.rb
+++ b/Casks/threema-multi-device-beta.rb
@@ -1,0 +1,24 @@
+cask "threema-multi-device-beta" do
+  arch arm: "arm64", intel: "x64"
+
+  version "2.0-beta17"
+  sha256 :no_check
+
+  url "https://releases.threema.ch/desktop/#{version}/threema-desktop-v#{version}-macos-#{arch}.dmg"
+  name "Threema Desktop 2.0 Beta"
+  desc "End-to-end encrypted instant messaging application"
+  homepage "https://threema.ch/download-md"
+
+  livecheck do
+    url "https://threema.ch/en/download-md"
+    regex(/(2.0-beta\s*\d+(?:)+)/i)
+  end
+
+  app "Threema Beta.app"
+
+  zap trash: [
+    "~/Library/Application Support/ThreemaDesktop",
+    "~/Library/Preferences/ch.threema.threema-desktop.plist",
+    "~/Library/Saved Application State/ch.threema.threema-desktop.savedState",
+  ]
+end
--- a/Casks/threema-beta.rb
+++ b/Casks/threema-beta.rb
@@ -6,13 +6,13 @@ cask "threema-beta" do
          intel: "1f20ae86c78c65be029e17c0e07a3c07cc9a780f4376f451f39fbdcd10f3ec48"
 
   url "https://releases.threema.ch/desktop/#{version}/threema-desktop-v#{version}-macos-#{arch}.dmg"
-  name "Threema Desktop 2.0 Beta"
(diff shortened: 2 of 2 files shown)


Casks/threema-multi-device-beta.rb
```diff
@@ -0,0 +1,24 @@
+cask "threema-multi-device-beta" do
```
bevanjkay, line 1: @joelfischerr The token should be `threema-beta` according to the [token reference guidelines.](https://docs.brew.sh/Cask-Cookbook#token-reference)

Casks/threema-multi-device-beta.rb
```diff
@@ -0,0 +1,24 @@
...
+  arch arm: "arm64", intel: "x64"
+
+  version "2.0-beta17"
+  sha256 :no_check
```
bevanjkay, line 5: @joelfischerr Because the url is versioned, we should have a shasum here instead of `:no_check`

Casks/threema-multi-device-beta.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  url "https://releases.threema.ch/desktop/#{version}/threema-desktop-v#{version}-macos-#{arch}.dmg"
+  name "Threema Desktop 2.0 Beta"
```
bevanjkay, line 8:
```suggestion
  name "Threema"
```

Casks/threema-multi-device-beta.rb
```diff
@@ -0,0 +1,24 @@
...
+  desc "End-to-end encrypted instant messaging application"
+  homepage "https://threema.ch/download-md"
+
+  livecheck do
+    url "https://threema.ch/en/download-md"
+    regex(/(2.0-beta\s*\d+(?:)+)/i)
+  end
```
bevanjkay, line 15:
```suggestion
  livecheck do
    url "https://threema.ch/en/download-md"
    regex(/href=.*?threema[._-]desktop[._-]v?(\d+(?:(?:[.-]|(beta))+\d+)+)[._-]macos[._-]arm64\.dmg/i)
  end
```

bevanjkay (changes requested):
@joelfischerr Please see requested changes

joelfischerr:
Thank you for the review!

I've made the requested changes. Your suggestions were not applied because I already pushed the first fixup commit and then Github wouldn't let me apply the now "outdated" suggestions.

bevanjkay:
@joelfischerr Thanks for the changes. Before I merge can you please confirm that this can be used alongside the stable version? If not we should add a `conflicts_with` stanza.

joelfischerr:
There is not stable version at the moment. Should we add the `conflicts_with` anyways to be sure nobody forgets to update it once the stable version is released?

bevanjkay:
@joelfischerr is this not the stable version?

https://github.com/Homebrew/homebrew-cask/blob/fa103d2c4550e4e3b2b7a4854681fc601a2cf393/Casks/threema.rb

joelfischerr:
I believe it is a different product using different identifiers. It doesn't
conflict with that one for sure.

bevanjkay:
@joelfischerr I would say that it is the beta for the second version of the same cask, but it looks like it doesn't share an preferences, so both can be installed concurrently.

bevanjkay (approved):

PR #17662, opened 2023-09-09
virtualbox-beta: add `livecheck`

miccal:
There may be a better way ...

### DIFF
--- a/Casks/virtualbox-beta.rb
+++ b/Casks/virtualbox-beta.rb
@@ -17,7 +17,7 @@ cask "virtualbox-beta" do
 
   livecheck do
     url :homepage
-    regex(/VirtualBox[._-]v?(\d+(?:.)+)[._-]#{arch}.dmg(?!.+development)/i)
+    regex(/href=.*?VirtualBox[._-]v?(\d+(?:[.-]\d+)+.*?)[._-]#{arch}\.dmg(?!.+?development)/i)
   end
 
   conflicts_with cask: "virtualbox"
--- a/Casks/virtualbox-beta.rb
+++ b/Casks/virtualbox-beta.rb
@@ -15,14 +15,9 @@ cask "virtualbox-beta" do
   desc "Virtualizer for x86 and arm64 hardware"
   homepage "https://www.virtualbox.org/wiki/Testbuilds"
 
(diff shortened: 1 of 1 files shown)


Casks/virtualbox-beta.rb
```diff
@@ -15,14 +15,9 @@
...
-  # TODO: Add a `livecheck` block if/when ARM64 dmg files are stored in a
-  # predictable directory on download.virtualbox.org. ARM64 files are
-  # currently shoehorned into stable version directories instead of the
-  # usual unstable version directories (e.g., 7.0.0_BETA3). At the moment,
-  # we would have to check both the highest unstable and stable version
-  # directories for an ARM64 file to find the newest version from each.
   livecheck do
-    skip "No reliable way to get version information"
+    url :homepage
+    regex(/VirtualBox[._-]v?(\d+(?:.)+)[._-]#{arch}.dmg(?!.+development)/i)
```
bevanjkay, line 20: @miccal Is the `.` in the main capture group intentionally unescaped here?

miccal: Yes, that was the easiest  way I could get it to match `7.0.11_BETA4-158798` and `7.0.11-158813`.

samford:
```suggestion
    regex(/href=.*?VirtualBox[._-]v?(\d+(?:\.\d+)+.*?)[._-]#{arch}\.dmg.+?development\s+revision/i)
```

Just some small tweaks but the general idea remains the same. I tested it out and it works as expected for both Intel and ARM (assuming we're targeting the "Development Snapshot" builds) 👍

miccal: Thanks @samford, but we are NOT targeting the "Development snapshots", just the "Latest 7.0.x test builds".

samford: Ah! Thanks for clarifying. I misread the regex and overlooked the _negative_ "development" lookahead (it's a bit late over here 😅). I've updated the suggestion above.

miccal: Thanks again @samford.

samford (approved):

PR #17976, opened 2023-10-09
defold-alpha: update `livecheck`

miccal:
Same as https://github.com/Homebrew/homebrew-cask/pull/157165

### DIFF
--- a/Casks/defold-alpha.rb
+++ b/Casks/defold-alpha.rb
@@ -10,10 +10,21 @@ cask "defold-alpha" do
   desc "Game engine for development of desktop, mobile and web games"
   homepage "https://defold.com/"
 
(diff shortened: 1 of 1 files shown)


samford (commented):
We generally use the `GithubReleases` strategy these days when we need to check multiple releases like this. [Since these casks use unstable versions (some of which are marked as "pre-release" on GitHub), these would need a `strategy` block to opt out of the strategy's default pre-release filtering.]

If you're on board with this, I can push a commit to this and the `defold-beta` PR to update the `livecheck` blocks to use the strategy.

miccal:
Please go ahead @samford., thanks.

p-linnane (approved):

PR #18093, opened 2023-10-18
android-studio-preview-canary: update `conflicts_with` and remove `depends_on`

miccal:
The `android-studio-preview-canary` and `android-studio-preview-beta` Casks often drop the `Preview` from the `app` name (see for example https://github.com/Homebrew/homebrew-cask-versions/pull/18069), so to be on the safe side I suggest we make all three `android-studio` Casks conflict with each other.

Also, the `LSMinimumSystemVersion` in the `plist` is macOS 10.8, so removing the `depends_on`.

### DIFF
--- a/Casks/android-studio-preview-canary.rb
+++ b/Casks/android-studio-preview-canary.rb
@@ -17,8 +17,10 @@ cask "android-studio-preview-canary" do
   end
 
   auto_updates true
-  conflicts_with cask: "android-studio-preview-beta"
-  depends_on macos: ">= :high_sierra"
+  conflicts_with cask: [
+    "android-studio",
+    "android-studio-preview-beta",
+  ]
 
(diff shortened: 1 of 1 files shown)


p-linnane (approved):

louis993546:
Is there another way to solve the `Casks often drop the Preview from the app name ` issue without making these 3 all mutually exclusive? Cause being able to switch between difference versions of AS is not that unusual  (at least IMO), e.g. [different project uses different Android Gradle Plugin version, which depends on different versions of Android Studio](https://developer.android.com/build/releases/gradle-plugin#android_gradle_plugin_and_android_studio_compatibility).

bevanjkay:
We could consider always renaming the preview version to `Android Studio Preview.app` using the `target` parameter.

PR #18095, opened 2023-10-19
README.md: Correct repository name

zmc:
`brew tap homebrew-cask/versions` prompts the user for github credentials - because that repo doesn't exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ project.
 After you install Homebrew, run the following command:
 
 ```sh
-brew tap homebrew/cask-versions
+brew tap Homebrew/homebrew-cask-versions
 ```
 
 You can now install alternate versions of Casks.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,7 @@ project.
 After you install Homebrew, run the following command:
 
 ```sh
-brew tap homebrew/cask-versions
+brew tap Homebrew/homebrew-cask-versions
```
bevanjkay, line 11: @zmc The `homebrew-` is usually omitted when running `brew tap`.

bevanjkay:
> `brew tap homebrew-cask/versions` prompts the user for github credentials - because that repo doesn't exist.

The command in the description here is incorrectly written. If you instead run `brew tap homebrew/cask-versions` does it succeed?

PR #18259, opened 2023-11-02
google-chrome-canary: switch to `version :latest`

miccal:
In the last week alone this Cask has been updated 12 times.

The `livecheck` `url` at https://chromiumdash.appspot.com/fetch_releases?channel=Canary&platform=Mac lists the current version and the previous version for each new release, which allows them to be compared.

I propose that we throttle the updates to every 5th release.

The `livecheck` may of course fail at the moment, but if other maintainers agree the `ci-skip-livecheck` label can be added to allow this to pass the CI checks if necessary.

Ping @samford for a review of my `livecheck` block and @reitermarkus who did something similar for the `bettertouchtool` Cask here: https://github.com/Homebrew/homebrew-cask/pull/142173

### DIFF
--- a/Casks/google-chrome-canary.rb
+++ b/Casks/google-chrome-canary.rb
@@ -13,7 +13,6 @@ cask "google-chrome-canary" do
   #   regex(/"version":\s*"v?(\d+(?:\.\d+)+)"/i)
   # end
 
-  auto_updates true
   depends_on macos: ">= :catalina"
 
   app "Google Chrome Canary.app"
--- a/Casks/google-chrome-canary.rb
+++ b/Casks/google-chrome-canary.rb
@@ -7,6 +7,9 @@ cask "google-chrome-canary" do
   desc "Web browser"
   homepage "https://www.google.com/chrome/canary/"
 
+  # Canary releases are frequent, so this `strategy` block throttles livecheck
+  # to every fifth release. Ideally this should be handled outside of livecheck
+  # but this is an interim workaround.
   livecheck do
     url "https://chromiumdash.appspot.com/fetch_releases?channel=Canary&platform=Mac"
     strategy :page_match do |page|
--- a/Casks/google-chrome-canary.rb
+++ b/Casks/google-chrome-canary.rb
@@ -9,7 +9,19 @@ cask "google-chrome-canary" do
 
   livecheck do
     url "https://chromiumdash.appspot.com/fetch_releases?channel=Canary&platform=Mac"
-    regex(/"version":\s*"v?(\d+(?:\.\d+)+)"/i)
+    strategy :page_match do |page|
+      current_version = page.match(/"version":\s*"v?(\d+(?:\.\d+)+)"/i)
+      previous_version = page.match(/"previous_version":\s*"v?(\d+(?:\.\d+)+)"/i)
+
+      # Throttle updates to every 5th release.
+      version = if current_version[1].tr(".", "").to_i >= previous_version[1].tr(".", "").to_i + 50
+        current_version[1]
+      else
+        previous_version[1]
+      end
+
+      version.to_s
+    end
   end
 
   auto_updates true
--- a/Casks/google-chrome-canary.rb
+++ b/Casks/google-chrome-canary.rb
@@ -12,18 +12,20 @@ cask "google-chrome-canary" do
   # but this is an interim workaround.
   livecheck do
     url "https://chromiumdash.appspot.com/fetch_releases?channel=Canary&platform=Mac"
-    strategy :page_match do |page|
-      current_version = page.match(/"version":\s*"v?(\d+(?:\.\d+)+)"/i)
-      previous_version = page.match(/"previous_version":\s*"v?(\d+(?:\.\d+)+)"/i)
+    strategy :json do |json|
+      versions = json.map { |item| item["version"] }
+                     .compact
+                     .uniq
+                     .sort_by { |v| Version.new(v) }
+                     .reverse
+      current_pos = versions.index(version)
 
-      # Throttle updates to every 5th release.
-      version = if current_version[1].tr(".", "").to_i >= previous_version[1].tr(".", "").to_i + 50
-        current_version[1]
-      else
-        previous_version[1]
-      end
+      # Fall back to the newest version if the current version isn't found
+      next versions.first if current_pos.blank?
 
-      version.to_s
+      # Return the newest version if there have been five or more releases
+      # since the version in the cask.
+      (current_pos >= 5) ? versions.first : version
     end
   end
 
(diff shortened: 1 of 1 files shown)


miccal:
Passing CI at the moment since:
```
|-> brew livecheck google-chrome-canary
google-chrome-canary: 121.0.6101.0 ==> 121.0.6102.0
```

bevanjkay:
This gets updated by upstream so frequently that it might be better to just go `version :latest`. 

Essentially every time you run `brew upgrade google-chrome-canary` or `brew upgrade --greedy` you would get a newer version. 

This assumes there is a static url that points to the latest version.

Casks/google-chrome-canary.rb
```diff
@@ -9,7 +9,19 @@
 
   livecheck do
```
samford, line 10:
```suggestion
  # Canary releases are frequent, so this `strategy` block throttles livecheck
  # to every fifth release. Ideally this should be handled outside of livecheck
  # but this is an interim workaround.
  livecheck do
```

Casks/google-chrome-canary.rb
```diff
@@ -9,7 +9,19 @@
...
+      previous_version = page.match(/"previous_version":\s*"v?(\d+(?:\.\d+)+)"/i)
+
+      # Throttle updates to every 5th release.
+      version = if current_version[1].tr(".", "").to_i >= previous_version[1].tr(".", "").to_i + 50
+        current_version[1]
+      else
+        previous_version[1]
+      end
+
+      version.to_s
```
samford, line 23:
```suggestion
    strategy :json do |json|
      versions = json.map { |item| item["version"] }
                     .compact
                     .uniq
                     .sort_by { |v| Version.new(v) }
                     .reverse
      current_pos = versions.index(version)

      # Fall back to the newest version if the current version isn't found
      next versions.first if current_pos.blank?

      # Return the newest version if there have been five or more releases
      # since the version in the cask.
      (current_pos >= 5) ? versions.first : version
```

samford (commented):
I don't have an opinion regarding `version :latest` here but if we want to restrict the `livecheck` block to every fifth release, we're going to need to properly work with the versions in the response. I've added a suggestion below that uses the `Json` strategy to reverse sort the items by version (i.e., newest to oldest) and conditionally return either the newest version or the cask version.

If the cask version isn't found in the JSON response, then the check returns the newest version as a fallback. When the cask version is present in the JSON response, the check returns the newest version if there have been 5+ releases since the cask version, otherwise it returns the cask version.

For example, the `versions` array is currently `["121.0.6102.0", "121.0.6101.0", "121.0.6100.0", "120.0.6099.2", "120.0.6099.0", "120.0.6098.0", "120.0.6097.0", "120.0.6096.0", "120.0.6095.0", "120.0.6094.0"]` and the cask version is `121.0.6101.0`, so it will return the cask version (i.e., `current_pos == 1`, which is less than 5). If the cask version is set to `120.0.6098.0` (or a version not in the array), the check will return the newest version (121.0.6102.0).

Is this what you're looking for here?

-----

Past that, I've added a comment above the `livecheck` block to explain the situation, as we should be documenting instances where the `livecheck` block is being used to throttle versions.

It goes without saying but I'm still of the opinion that livecheck should always return the newest version and any throttling should be handled outside of it (e.g., failing fast on CI, preventing `bump-cask-pr` from creating a new PR, etc.). However, as mentioned before, I'm okay with this as an interim solution to conserve maintainer (and CI) time with respect to frequently-updated casks but someone should eventually implement a proper solution (e.g., homebrew/core has version throttling via an `audit_exceptions/throttled_formulae.json` file).

razvanazamfirei:
> It goes without saying but I'm still of the opinion that livecheck should always return the newest version and any throttling should be handled outside of it (e.g., failing fast on CI, preventing `bump-cask-pr` from creating a new PR, etc.). However, as mentioned before, I'm okay with this as an interim solution to conserve maintainer (and CI) time with

Fully agree with @samford on this. We implemented something similar in https://github.com/Homebrew/homebrew-cask/pull/145477 and https://github.com/Homebrew/homebrew-cask/pull/144441.

Not sure if it makes more sense to stick to a date-based throttle versus one based on the version number.

p-linnane:
Every 5 releases or `version :latest` work for me. Whatever more maintainers think makes sense.

bevanjkay:
I am leaning heavily toward `version :latest` for this one for a couple of reasons. It is a nightly build, that sometimes releases additional patches (sometimes multiple times per day).
It automatically updates. With `version :latest` users can still upgrade by calling the cask directly or using `--greedy` and due to the release schedule it would almost always upgrade following the shasum comparison check.

It is also worth noting the the `url` for this cask is **not** versioned, so there is no great advantages by having the cask versioned, as you may receive a newer version anyway.

miccal:
Thank you for the review @samford.

I think this should remain versioned to maintain consistency with the other three `google-chrome` Casks (which all have non-versioned `url`'s).

I propose we give throttling a try and see how it goes.

Casks/google-chrome-canary.rb
miccal: I though my way was clever 😢

samford: It's not that it wasn't clever, it's that the numeric approach wouldn't strictly throttle to every five versions.

Using the current releases as an example:

* Eight change the third number (e.g., 120.0.6094.0 -> 120.0.6095.0)
* One changes the fourth number but not the third number (120.0.6099.0 -> 120.0.6099.2)
* One changes the first, third, and fourth numbers (120.0.6099.2 -> 121.0.6100.0)

Releases that only modify the fourth number but not the third number wouldn't be counted as a release when using the numeric approach. Also, releases that bump the major version would always make `livecheck` return a new version, regardless of the throttling (though one could argue that's a feature 😆).

Due to the variation in what part of the version changes, strictly throttling to every fifth release ends up being more involved (as seen).

samford:
I was ambivalent about which `version` approach to take but I think Bevan's explanation has convinced me about `version :latest`. If Chrome updates itself and use of `--greedy` isn't super commonplace (I don't use it 🤷), then it's just about 1) ensuring that users get the latest version when they install the cask and 2) making sure that whatever we do reduces the current maintenance burden.

If we throttle this cask, 4/5 of the time users wouldn't be installing the latest version and they may be prompted to update (in Chrome) right after installation, which isn't ideal. Throttling would reduce the maintainer burden but lead to a slightly worse situation for users.

At least to me, `version :latest` seems to be suitable from both a user and maintainer standpoint, considering this is basically a "nightly" cask and the `url` is unversioned (as Bevan mentioned). That said, I'm not a cask expert by any means, so take this with a grain of salt.

miccal:
Switching to `version :latest`.

bevanjkay (approved):

PR #18374, opened 2023-11-14
pdf-expert-beta: update `depends_on` and `zap`

miccal:


Casks/pdf-expert-beta.rb
```diff
@@ -14,12 +14,12 @@
...
   auto_updates true
   conflicts_with cask: "pdf-expert"
-  depends_on macos: ">= :high_sierra"
+  depends_on macos: ">= :monterey"
 
   app "PDF Expert.app"
 
   zap trash: [
-    "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.readdle.pdfexpert-mac.sfl2",
+    "~/Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.ApplicationRecentDocuments/com.readdle.pdfexpert-mac.sfl*",
```
bevanjkay, line 22: @miccal If we are concerned about Apple changing the filename for the shared file list would it be better to glob the whole filename and not just the `2`?

miccal: Only the last digit changes between macOS versions and installs; it can be none, 1, 2 or 3 (to date).

miccal: Old issue: https://github.com/Homebrew/homebrew-cask/issues/40557

bevanjkay (approved):

PR #18513, opened 2023-11-28
anki-beta: remove

samford:

- [ ] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [ ] `brew install --cask <cask>` worked successfully.
- [ ] `brew uninstall --cask <cask>` worked successfully.

-----

~The `anki-beta` `livecheck` block uses a GitHub releases URL with the `GitHubReleases` strategy but it should simply use `url :url`. This PR updates the `livecheck` block accordingly.~

This removes the `anki-beta` cask, as upstream removes unstable versions when a subsequent stable version is released. With that setup, the cask will break if it's using an unstable version and a new stable version is released, so this isn't something that we can reasonably maintain (see the discussion below).


samford:
Hm, it looks like upstream deletes unstable releases on GitHub once the stable version appears:

```
curl: (22) The requested URL returned error: 404
Error: Download failed on Cask 'anki-beta' with message: Download failed: https://github.com/ankitects/anki/releases/download/23.10.1rc3/anki-23.10.1-mac-apple-qt6.dmg
```

Should we update this cask to stable releases (and modify the `livecheck` block to also surface those versions)? If not, this is probably going to be broken a good amount of the time.

p-linnane:
I don't really see any other way to handle it than what you suggested.

bevanjkay:
@samford We could use this directory instead?
https://apps.ankiweb.net/downloads/beta/?C=N&O=D
Linked from here: https://betas.ankiweb.net/

samford:
> We could use this directory instead?
> https://apps.ankiweb.net/downloads/beta/?C=N&O=D
> Linked from here: https://betas.ankiweb.net/

`anki-beta` originally checked that URL (https://github.com/Homebrew/homebrew-cask-versions/pull/13663) but the newest version on that page is 23.10 whereas the newest unstable version is now 23.12beta1 on GitHub (as of 13 hours ago). We also don't want to be checking that first-party page for version information when the cask is using a release asset from GitHub (especially with upstream deleting unstable releases from GitHub).

I'll update this to bump the version and rework the `livecheck` block (to also match stable versions) in a bit.

bevanjkay:
@samford This is a frustrating one, because the cask is going to break every time a release is cut because the previous binary has been deleted. As far as I know we generally aren't supporting casks in this situation. 

I'm leaning toward this being more suitable for removal, especially with pending migrating these casks to `homebrew-casks` where changes are only live when the API is refreshed.

samford:
> This is a frustrating one, because the cask is going to break every time a release is cut because the previous binary has been deleted. As far as I know we generally aren't supporting casks in this situation.

Looking back at archived snapshots of the releases page (like the one linked above), all the unstable versions leading up to the stable release exist until the stable version is released. A series of unstable releases would be fine but there's no getting around the fact that the cask will break with a new stable release (until it's updated). I agree that it's not something that we can reasonably maintain.

> I'm leaning toward this being more suitable for removal, especially with pending migrating these casks to `homebrew-casks` where changes are only live when the API is refreshed.

`anki-beta` has only had two installs in the past 90 days (zero in the past 30), so removal seems fine to me. The few people who want this cask are free to make a copy and maintain it in their own tap. I'll update this to remove the cask momentarily.

bevanjkay (approved):

PR #19124, opened 2024-01-29
Add zed-preview cask

timfjord:
A preview version of the Zed editor - https://zed.dev/releases/preview

- [x] `brew audit --cask --new <cask>` worked successfully.
- [x] `HOMEBREW_NO_INSTALL_FROM_API=1 brew install --cask <cask>` worked successfully.

### DIFF
--- a/Casks/zed-preview.rb
+++ b/Casks/zed-preview.rb
@@ -13,9 +13,7 @@ cask "zed-preview" do
   end
 
   auto_updates true
-  conflicts_with cask: [
-    "zed",
-  ]
+  conflicts_with cask: "zed"
   depends_on macos: ">= :catalina"
 
   app "Zed Preview.app"
--- a/Casks/zed-preview.rb
+++ b/Casks/zed-preview.rb
@@ -1,0 +1,30 @@
+cask "zed-preview" do
+  version "0.120.3"
+  sha256 "2c44a02b7667a324933e779488c3c70c949089741ee387b39cda463dd3f3558e"
+
+  url "https://zed.dev/api/releases/preview/#{version}/Zed.dmg"
+  name "Zed Preview"
+  desc "Multiplayer code editor"
+  homepage "https://zed.dev/"
+
+  livecheck do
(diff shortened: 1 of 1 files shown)


Casks/zed-preview.rb
```diff
@@ -0,0 +1,30 @@
...
+  desc "Multiplayer code editor"
+  homepage "https://zed.dev/"
+
+  livecheck do
+    url "https://zed.dev/releases/preview"
+    regex(%r{href=.*?/preview/(\d+(?:\.\d+)+)/Zed.dmg}i)
+  end
+
+  auto_updates true
+  conflicts_with cask: [
```
timfjord, line 16: Regarding this one.
It doesn't really conflict with the stable release version, because the name of app is `Zed Preview.app`.
So we could change the `binary` to be something like
```ruby
binary "#{appdir}/Zed Preview.app/Contents/MacOS/cli", target: "zed-preview"
```
and remove the `conflicts_with`.

But the problem here is the app doesn't use `zed-preview` for the binary, so if someone tries to link from the app they will have two binaries for the same installation.
So I would probably keep the `conflicts_with` to avoid confusion and to follow the app defaults.

timfjord: And yes, if we agree on that I will go ahead and add the `conflicts_with` to the stable cask too.

bevanjkay (changes requested):
Just one small formatting nitpick that unfortunately does not get picked up by `brew style` at the moment.

Otherwise looks great, and I agree with your assessment of the conflict.

Casks/zed-preview.rb
```diff
@@ -0,0 +1,30 @@
...
+    "zed",
+  ]
```
bevanjkay, line 18:
```suggestion
  conflicts_with cask: "zed"
```

Just a small formatting nitpick here, that doesn't currently get picked up by

bevanjkay (approved):

krehel (approved):

PR #19300, opened 2024-02-13
autobump: autobump more casks every six hours

chenrui333:
- autobump: autobump more casks every six hours

### DIFF
--- a/.github/workflows/autobump.yml
+++ b/.github/workflows/autobump.yml
@@ -13,33 +13,105 @@ on:
 env:
   CASKS: >
     1password-beta
+    1password-cli1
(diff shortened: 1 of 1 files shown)


bevanjkay:
I don't think auto bump runs in this repo

chenrui333:
> I don't think auto bump runs in this repo

the action runs are in here (https://github.com/Homebrew/homebrew-cask-versions/actions/workflows/autobump.yml)?

bevanjkay:
There's no token so it never actually completes a bump. I think the plan was to just defer until the repo is merged with `homebrew-cask`

PR #19495, opened 2024-02-28
wine-devel 9.3

chenrui333:
Created by `brew bump`

---


YupItzAfi:
Should add this👍

YupItzAfi (commented):
gstreamer issue is with homebrew, not with wine-devel

p-linnane:
Tried the same approach that worked for https://github.com/Homebrew/homebrew-cask-versions/pull/19493, but there are only 3 binaries in the `bin` folder for this version. I'm not sure if that is intentional or not.

YupItzAfi:
Then why would they remove winedbg winecfg etc too...

p-linnane:
Thank you very much @Gcenx. Your help is very appreciated.

YupItzAfi:
@Gcenx sorry, my bad, thanks either way!

p-linnane (approved):

PR #19586, opened 2024-03-07
safari-technology-preview 190

hannesbraun:


samford:
Looking at the homepage, the `livecheck` block is failing on macOS 12 and older because the regex needs to match the "Requires macOS 13 or later" text but it's looking for "macOS 12", etc. instead.

We can address this by either A) manually specifying Ventura's macOS version string for the regex in the `on_ventura :or_older` block (i.e., Monteray would use "13") or B) strictly using `on_ventura` (with no `:or_older`). The "Requires macOS 13 or later" text on the homepage suggests that older versions aren't supported.

If we want to maintain the current `on_ventura :or_older` condition, I can push a `livecheck` block (I already have it worked out and tested). Otherwise, if the "Requires macOS 13 or later" text is correct, I can push a commit to update the `on_system` block.

Thoughts on this?

p-linnane:
Safari Technology Preview only supports the current and previous version of macOS. So we should limit this to Ventura + Sonoma. We'll have to update it in the fall to drop Ventura and add macOS 15.

samford:
Thanks for confirming, @p-linnane. I pushed a commit to remove the `:or_older` argument from the `on_ventura` call, so the `on_system` calls align with `depends_on macos: ">= :ventura"`.

However, this will predictably fail on macOS versions older than Ventura, where `version` isn't set:

```
Error: Invalid cask (macOS 12 on intel): /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask-versions/Casks/safari-technology-preview.rb
  Cask 'safari-technology-preview' is unreadable: undefined method `csv' for nil:NilClass
```

Is there a standard/existing approach for how to handle this scenario in casks?

p-linnane:
Thanks! There is not a standard approach to my knowledge, but CC @Homebrew/cask for any other input.

bevanjkay:
I think scoping the `livecheck` within the specific blocks is the best option here.
It's a fairly manual update anyway as the scoping changes when there's MacOS version changes, so also adjusting the `livecheck` does not add considerable time.

miccal (approved):

PR #20115, opened 2024-04-22
firefox-*: set auto_updates to true

scientificworld:


bevanjkay (approved):

scientificworld:
> Ahh.. it looks like because these are all "unversioned" casks (set to `version :latest`), that `auto_updates true` does not apply, as the logic conflicts, with `version :latest` taking precedence.

Oh sorry... I just found out that only firefox-esr need to set this. Can I open a new PR to fix it?

bevanjkay:
Sure, sounds good 👍🏻

